PR #7, opened 2015-06-26
add two convenience methods for handling output

### LINKED ISSUE #4: Error helper
Most of my CLI's have this:

```js
var cli = meow();

fn(function (err) {
	if (err) {
		console.error(err.message);
		process.exit(1);
	}
});
```

Wonder if we could abstract out the error handling.

Eg:

```js
var cli = meow();

fn('unicorn', function (err) {
	cli.handleError(err);
});
```

It could potentially also handle showing the stack when `--verbose` is passed.

Thoughts?

-

// @kevva @floatdrop @arthurvr @shinnn

sindresorhus:
**NOT READY TO BE MERGED**

fixes #4

-

Just a quick change to get discussion going.

This would handle the most common CLI cases for me.

Wonder if we could just go with one method though, that just logged if the second argument where not undefined?

Also not happy with the method names. Any better suggestions?

// @kevva @arthurvr @shinnn

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,20 @@ var camelcaseKeys = require('camelcase-keys');
 delete require.cache[__filename];
 var parentDir = path.dirname(module.parent.filename);
 
+function handleError(err) {
+	if (!err) {
+		return;
+	}
+
+	console.error(err.message);
+	process.exit(1);
+}
+
+function handleOutput(err, output) {
+	handleError(err);
+	console.log(output);
+}
+
 module.exports = function (opts, minimistOpts) {
 	opts = objectAssign({
 		pkg: './package.json',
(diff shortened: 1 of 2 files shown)


kevva:
Haven't had a chance to check the code thoroughly but a naming suggestion. Maybe `onError` for the error one. The other one is harder.

ruffle1986:
```catch``` instead of ```handleError``` ?

File: index.js [hunk 1]
```diff
@@ -9,6 +9,20 @@ var camelcaseKeys = require('camelcase-keys');
 delete require.cache[__filename];
 var parentDir = path.dirname(module.parent.filename);
 
+function handleError(err) {
+	if (!err) {
+		return;
+	}
+
+	console.error(err.message);
```
sindresorhus: @kevva I'm not convinced we should be doing this. It makes it really hard to debug as it hides the stack trace. Maybe throw if the `DEBUG` environment variable is set?

sindresorhus:
@ruffle1986 `catch` is a reserved keyword, so can't use that. Could maybe go with `error`.

sindresorhus:
Skipping this as we now handle promise errors automagically and I'm not going to continue using callback-style interface.

PR #10, opened 2015-07-09
Exit code argument for showHelp

eush77:
Printing usage text in response to `--help` argument is a normal execution but showing help and terminating the app for other reasons is often not.

For example, on my system `sed --help` shows help for `sed` and exits with code 0, but `sed -x` prints

```
sed: invalid option -- 'x'
```

followed by (mostly) the same usage text and exits with non-zero code 4. However, there is currently no way to implement this behavior using meow's `showHelp` method.

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,9 +26,9 @@ module.exports = function (opts, minimistOpts) {
 	var pkg = typeof opts.pkg === 'string' ? require(path.join(parentDir, opts.pkg)) : opts.pkg;
 	var argv = minimist(opts.argv, minimistOpts);
 	var help = '\n' + indentString(pkg.description + (opts.help ? '\n\n' + opts.help : '\n'), '  ');
-	var showHelp = function () {
+	var showHelp = function (code) {
 		console.log(help);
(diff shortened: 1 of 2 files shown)


sindresorhus:
Node 0.12 has a [`process.exitCode`](https://nodejs.org/api/process.html#process_process_exitcode) setter ([polyfill](https://github.com/isaacs/exit-code)). I wonder if it would be better to just use that instead?

Relevant: https://github.com/sindresorhus/meow/issues/4#issuecomment-119185677

eush77:
Personally I like the explicitness and brevity of

```
cli.showHelp(1);
```

but this problem can indeed be solved with current API. Feel free to close this PR then.

sindresorhus:
I'll keep it open. Need to think about this.

sindresorhus:
Landed: https://github.com/sindresorhus/meow/commit/03c4a0f8f5077a1466432bd138f95b2be0572809

eush77:
Woo-hoo!

Issue #12, opened 2015-10-29
Outputs "undefined" if there is no description


callumlocke:
I don't want to output my package.json description at the top of my CLI's help text. There's no obvious way to disable this.

I tried passing in a custom pkg object, omitting the description field:

```js
import meow from 'meow';
import pkg from '../package.json';

const cli = meow(`
  Blah blah blah.
`, {
  pkg: {version: pkg.version}
});
```

...but this results in:

```sh
$ foo --help
  undefined

  Blah blah blah.
```

I guess I could set the description to `""` but that feels like a hack and leaves an extra newline.

This is a nice little library btw!

sindresorhus:
Why would you not want to show the description? It's a nice UX for the users.

callumlocke:
Sure, but I want to write my own description inside my help text, not necessarily identical to the one in package.json.

The pkg description is written for things like npm search results. If someone has already installed the app and is asking for help on the command line, it's a different context imo and calls for a more practical introduction sentence. In some cases anyway.

callumlocke:
(I agree inserting the pkg description is a sensible default though)

sindresorhus:
That's a good point. Didn't think of that use-case.

PR #13, opened 2015-10-29
Don't output pkg description if undefined

callumlocke:
See #12.


sindresorhus:
I don't think having to pass a modified `pkg` object is a good solution. Would be better to add a `description` option that can be either a custom description or `false` to disable, like the `version` option.

kevva:
> Would be better to add a description option that can be either a custom description or false to disable, like the version option.

:+1:

callumlocke:
I agree, I'll update the PR

callumlocke:
closing this in favour of #14

PR #14, opened 2015-10-29
Add description option

### LINKED ISSUE #12: Outputs "undefined" if there is no description

callumlocke:
fixes #12

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -79,6 +79,15 @@ Type: `object`, `array`, `string`
 
 Can either be a string/array that is the `help` or an options object.
 
+##### description
+
+Type: `string`, `boolean`
+Default: The package.json `"description"` property
+
+A description to show above the help text.
+
+Set it to `false` to disable it altogether.
+
 ##### help
 
 Type: `string`, `boolean`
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -79,6 +79,15 @@ Type: `object`, `array`, `string`
 
 Can either be a string/array that is the `help` or an options object.
 
+##### description
+
+Type: `string`, `boolean`
+Default: The package.json `"description"` property
+
+A description to show at the top of the help text.
```
sindresorhus: `at the top of`  → `above`

sindresorhus:
Looks good. Can you add a test?

sindresorhus:
Perfect.

PR #15, opened 2015-11-07
Add shortcut for resolving relative paths

denysdovhan:
I like meow and want to make it better.

In my own cli apps I often write `cwd`-method that resolves relative filenames and current working dir. So, I propose to add this feature into **meow**.

I've already written tests and desc for `cwd`-method, so please check my changes and merge if it'll be useful for everyone.

### DIFF
--- a/index.js
+++ b/index.js
@@ -52,6 +52,10 @@ module.exports = function (opts, minimistOpts) {
 		process.exit(code || 0);
 	};
 
+	var cwd = function (filename) {
+		return path.resolve(process.cwd(), filename || '');
+	};
+
 	if (argv.version && opts.version !== false) {
 		console.log(typeof opts.version === 'string' ? opts.version : pkg.version);
 		process.exit();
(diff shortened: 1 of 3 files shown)


sindresorhus:
Can you provide an example of when you would need something like this?

denysdovhan:
Of course.
[Here is example from textr-cli](https://github.com/denysdovhan/textr-cli/blob/dev/cli/index.js#L46-L52), that reads stdin or specified filename

sindresorhus:
All those uses are moot, though. You can pass relative paths directly to any core method, including the `fs` methods. No need to resolve it first.

denysdovhan:
Hmm. Sometimes I have problems with relative paths. Maybe it only my fault.

sindresorhus:
Yeah, I'm gonna pass on this. I very seldom have the need to resolve my paths and when I do, it's only `path.resolve('.', filename)`.

denysdovhan:
Okay, thanks.

PR #19, opened 2015-12-10
add type inference option

### LINKED ISSUE #17: Turn off cli args type detection by default
Can be turned on in an option.

I've had so many problems with it.

The latest: https://github.com/chalk/chalk-cli/pull/2

What should the option be named?

-

Note to self: Remove all instances of `string: ['_']` in my modules when this is done.

SamVerschueren:
Adds the option `inferType` to the options object. It's `false` by default which means that command line arguments are not parsed. (fixes #17)

Also fixed tests with latest AVA.

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,8 +25,13 @@ module.exports = function (opts, minimistOpts) {
 			cwd: parentDir,
 			normalize: false
 		}).pkg,
-		argv: process.argv.slice(2)
+		argv: process.argv.slice(2),
+		inferType: false
 	}, opts);
+
+	if (opts.inferType === false) {
+		minimistOpts = objectAssign(minimistOpts || {}, {string: ['_']});
+	}
 
 	if (Array.isArray(opts.help)) {
 		opts.help = opts.help.join('\n');
--- a/index.js
+++ b/index.js
@@ -31,8 +31,12 @@ module.exports = function (opts, minimistOpts) {
 
 	minimistOpts = objectAssign({string: ['_']}, minimistOpts);
 
-	if (opts.inferType === false && minimistOpts.string.indexOf('_') === -1) {
+	var index = minimistOpts.string.indexOf('_');
+
+	if (opts.inferType === false && index === -1) {
 		minimistOpts.string.push('_');
+	} else if (opts.inferType === true && index !== -1) {
+		minimistOpts.string.splice(index, 1);
 	}
 
 	if (Array.isArray(opts.help)) {
--- a/index.js
+++ b/index.js
@@ -29,8 +29,10 @@ module.exports = function (opts, minimistOpts) {
 		inferType: false
 	}, opts);
 
-	if (opts.inferType === false) {
-		minimistOpts = objectAssign(minimistOpts || {}, {string: ['_']});
+	minimistOpts = objectAssign({string: ['_']}, minimistOpts);
+
+	if (opts.inferType === false && minimistOpts.string.indexOf('_') === -1) {
+		minimistOpts.string.push('_');
 	}
 
 	if (Array.isArray(opts.help)) {
--- a/package.json
+++ b/package.json
@@ -49,7 +49,13 @@
   },
   "devDependencies": {
     "ava": "*",
+    "execa": "^0.1.1",
     "indent-string": "^2.1.0",
     "xo": "*"
+  },
+  "xo": {
+    "ignores": [
+      "test.js"
+    ]
   }
 }
--- a/readme.md
+++ b/readme.md
@@ -131,7 +131,15 @@ Custom arguments object.
 Type: `boolean`  
 Default: `false`
 
-Set it to `true` to enable type inference.
+Set it to `true` to enable argument type inference.
+
+###### example
+
+```
+$ ./foo-app.js 5
+```
+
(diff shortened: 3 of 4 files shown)


test.js
```diff
@@ -21,46 +23,29 @@ test('return object', t => {
...
 test('support help shortcut', t => {
 	const cli = fn(['unicorn', 'cat']);
 	t.is(cli.help, indentString('\nCLI app helper\n\nunicorn\ncat', '  '));
-	t.end();
 });
 
-test('spawn cli and show version', t => {
-	t.plan(2);
+test('spawn cli and show version', async t => {
+	const {stdout} = await execa('./fixture.js', ['--version'], {cwd: __dirname});
```
sindresorhus: You don't have to set `cwd`. It's always `__dirname`.

readme.md
```diff
@@ -126,6 +126,13 @@ Default: `process.argv.slice(2)`
 
 Custom arguments object.
 
+##### inferType
+
+Type: `boolean`  
+Default: `false`
+
+Set it to `true` to enable type inference.
```
sindresorhus: > Type inference.

Maybe also include an example of what `type inference` means in this context.

index.js
```diff
@@ -25,9 +25,14 @@ module.exports = function (opts, minimistOpts) {
 			cwd: parentDir,
 			normalize: false
 		}).pkg,
-		argv: process.argv.slice(2)
+		argv: process.argv.slice(2),
+		inferType: false
 	}, opts);
 
+	if (opts.inferType === false) {
+		minimistOpts = objectAssign(minimistOpts || {}, {string: ['_']});
```
sindresorhus: This will override any `string` config the user does. We should instead make it foolproof by ensuring `string` always contains `'_'`.

test.js
```diff
@@ -1,6 +1,8 @@
 import childProcess from 'child_process';
 import test from 'ava';
 import indentString from 'indent-string';
+import execa from 'execa';
+import {version as pkgVersion} from './package.json';
 import fn from './';
```
sindresorhus:
```js
global.Promise = Promise;
```

to fix the tests

test.js
```diff
@@ -1,6 +1,8 @@
 import childProcess from 'child_process';
 import test from 'ava';
 import indentString from 'indent-string';
+import execa from 'execa';
+import {version as pkgVersion} from './package.json';
```
sindresorhus:
```js
import pkg from './package.json';
```

is cleaner

index.js
SamVerschueren: Good point!

readme.md
```diff
@@ -126,6 +126,21 @@ Default: `process.argv.slice(2)`
...
 Custom arguments object.
 
+##### inferType
+
+Type: `boolean`  
+Default: `false`
+
+Set it to `true` to enable argument type inference.
+
+###### example
```
SamVerschueren: Not sure about the example.

test.js
```diff
@@ -72,10 +59,16 @@ test.serial('pkg.bin as a string should work', t => {
...
 	t.ok(fn({argv: ['-F']}).flags.F);
-	t.end();
+});
+
+test('type inference', t => {
+	t.is(fn({argv: ['5']}).input[0], '5');
+	t.is(fn({argv: ['5']}, {string: ['_']}).input[0], '5');
+	t.is(fn({argv: ['5'], inferType: true}).input[0], 5);
+	t.is(fn({argv: ['5'], inferType: true}, {string: ['foo']}).input[0], 5);
+	t.is(fn({argv: ['5'], inferType: true}, {string: ['_', 'foo']}).input[0], 5);
```
SamVerschueren: Is this correct? What if the user sets `inferType` to `true` but passes `_` in the string property? In my opinion this is correct as `inferType` has "higher priority" then the `string` property.

readme.md
SamVerschueren: What do you mean with the `Type inference.` note? Should I capitalize that?

> Set it to `true` to enable Type inference.

test.js
sindresorhus: Yup, options always takes precedence.

package.json
```diff
@@ -49,7 +49,13 @@
...
   "devDependencies": {
     "ava": "*",
+    "execa": "^0.1.1",
     "indent-string": "^2.1.0",
     "xo": "*"
+  },
+  "xo": {
+    "ignores": [
+      "test.js"
+    ]
```
sindresorhus: This is no longer needed.

readme.md
```diff
@@ -126,6 +126,21 @@ Default: `process.argv.slice(2)`
 
 Custom arguments object.
 
+##### inferType
+
+Type: `boolean`  
+Default: `false`
+
+Set it to `true` to enable argument type inference.
```
sindresorhus: > Infer the argument type.

sindresorhus:
Thank you! And sorry about the delay :)

SamVerschueren:
No problem, you have so much stuff to do :)

williamboman:
Any ETA on new release? Was reading the docs and could not figure out why it would keep inferring types. Also, wouldn't this break BC if it defaults to `false`?

PR #22, opened 2015-12-15
Add stdin and input handling

### LINKED ISSUE #20: stdin handling
Note, not reading, but handling when to read from stdin.

Currently it's this logic: https://github.com/sindresorhus/find-versions-cli/blob/56dbd77051147819abdf845504af3b3cc1dd2935/cli.js#L30-L41

The logic is, we only try to read from stdin when there's no input and `process.isTTY === false`. Otherwise it won't work when used with child_process.execFile so the logic is hard to get right for many
if you only check isTTY, which i did with most of my CLIs previously, it wouldn't work with execFile, as isTTY is false there and it tries to read from stdin.

SamVerschueren:
fixes #20 

Feedback is more then welcome.

Note: PR is made from branch iss17.

### DIFF
--- a/fixture.js
+++ b/fixture.js
@@ -16,3 +16,14 @@ var cli = meow({
 Object.keys(cli.flags).forEach(function (el) {
 	console.log(el);
 });
+
+cli.stdinOrInput()
+	.then(function (input) {
+		input.forEach(function (element) {
+			console.log(element);
+		});
+	})
+	.catch(function (err) {
+		console.error(err.message);
+		process.exit(1);
+	});
--- a/index.js
+++ b/index.js
@@ -8,10 +8,14 @@ var redent = require('redent');
 var readPkgUp = require('read-pkg-up');
 var loudRejection = require('loud-rejection');
 var normalizePackageData = require('normalize-package-data');
+var getStdin = require('get-stdin');
+var Promise = require('pinkie-promise');
 
 // get the uncached parent
 delete require.cache[__filename];
 var parentDir = path.dirname(module.parent.filename);
+
+global.Promise = Promise;
 
 module.exports = function (opts, minimistOpts) {
 	loudRejection();
@@ -25,8 +29,19 @@ module.exports = function (opts, minimistOpts) {
 			cwd: parentDir,
 			normalize: false
 		}).pkg,
-		argv: process.argv.slice(2)
+		argv: process.argv.slice(2),
+		inferType: false
 	}, opts);
+
+	minimistOpts = objectAssign({string: ['_']}, minimistOpts);
+
+	var index = minimistOpts.string.indexOf('_');
+
+	if (opts.inferType === false && index === -1) {
+		minimistOpts.string.push('_');
+	} else if (opts.inferType === true && index !== -1) {
+		minimistOpts.string.splice(index, 1);
+	}
 
 	if (Array.isArray(opts.help)) {
 		opts.help = opts.help.join('\n');
@@ -69,6 +84,19 @@ module.exports = function (opts, minimistOpts) {
 		flags: camelcaseKeys(argv),
 		pkg: pkg,
 		help: help,
-		showHelp: showHelp
+		showHelp: showHelp,
+		stdinOrInput: function () {
+			if (!_ && process.stdin.isTTY) {
+				return Promise.reject(new Error('input required'));
+			}
+
+			if (_.length > 0) {
+				return Promise.resolve(_);
+			}
+
+			return getStdin().then(function (data) {
+				return [data];
+			});
+		}
 	};
 };
--- a/test.js
+++ b/test.js
@@ -1,7 +1,10 @@
-import childProcess from 'child_process';
 import test from 'ava';
 import indentString from 'indent-string';
+import execa from 'execa';
+import pkg from './package.json';
 import fn from './';
+
+global.Promise = Promise;
 
 test('return object', t => {
 	const cli = fn({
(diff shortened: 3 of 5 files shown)


SamVerschueren:
Will fix the tests tomorrow.

sindresorhus:
Can you rebase when you have a chance?

SamVerschueren:
I might have done something wrong. I'm not able to squash those commits. The first commit also has the changes of the previous PR (#19 - add type inference option) which I merged from master. You're probably way better in it then me :).

SamVerschueren:
@sindresorhus ping

I know you're quite busy lately, just reply when you find the time.

index.js
```diff
@@ -80,6 +84,19 @@ module.exports = function (opts, minimistOpts) {
 		flags: camelcaseKeys(argv),
 		pkg: pkg,
 		help: help,
-		showHelp: showHelp
+		showHelp: showHelp,
+		stdinOrInput: function () {
+			if (!_ && process.stdin.isTTY) {
+				return Promise.reject(new Error('input required'));
```
sindresorhus: `input` => `Input`

sindresorhus:
There are two more use-cases. Just getting stdin without input, and getting stdin or file. Maybe we should also add: `.stdin()` (just exposing `get-stdin`) and `.stdinOrFile()` (which would `fs.readFileSync` the input). Any ideas on how could simplify this. Not sure how I feel about adding three methods. Are we taking the right approach here?

fixture.js
```diff
@@ -16,3 +16,14 @@ var cli = meow({
 Object.keys(cli.flags).forEach(function (el) {
 	console.log(el);
 });
+
+cli.stdinOrInput()
+	.then(function (input) {
+		input.forEach(function (element) {
```
sindresorhus: The main use-case is one input argument, so would really like to optimize for that, but still support the edge-case of multiple arguments.

SamVerschueren:
> Any ideas on how could simplify this. Not sure how I feel about adding three methods. Are we taking the right approach here?

Good point, it makes the module more complex. Not sure what you think about a wrapper like `meow-stdin` which could wrap `meow` and add all these extra methods.

```js
const meowStdin = require('meow-stdin');
const meow = meowStdin(require('meow'));

// and son on
```

Or a separate package that uses meow in the background but adds `stdin` methods.

```js
const meow = require('meow-stdin');

// and son on
```

test.js
```diff
@@ -71,3 +71,11 @@ test('type inference', t => {
 	t.is(fn({argv: ['5'], inferType: true}, {string: ['foo']}).input[0], 5);
 	t.is(fn({argv: ['5'], inferType: true}, {string: ['_', 'foo']}).input[0], 5);
 });
+
+test('stdinOrInput', async t => {
+	const stdin = await execa.shell('echo \'foo bar\' | ./fixture.js');
+	const input = await execa('./fixture.js', ['foo', 'bar']);
+
+	t.is(stdin.stdout, 'meow\nfoo bar\n');
+	t.is(input.stdout, 'meow\nfoo\nbar');
```
Qix-: This isn't how stdin is intended. At least, this is completely counterintuitive for any other program out there. Usually how it works is `program file.txt` or `cat file.txt | program

SamVerschueren: Is there a big difference in `echo "foo bar" | cli` and `cat file.txt | cli` where the content of `file.txt` is `foo bar`? I assumed, if I test `echo "foo bar" | cli`, then `cat file.txt | cli` will work as well?

Qix-: That's not what I mean. Usually programs that accept _filenames_ as arguments are the ones that employ the use of `stdin`.

For instance:

```shell
grep "text" some-file.txt
echo "some text" | grep "text"

sed 's/some/thing/' some-file.txt
echo "some text" | sed 's/some/thing/'
```

Instead, this PR seems to support the usage of:

```shell
cat > some-prog.js <<EOF
// ...
console.log(meow.input);
// ...
EOF

some-prog.js This Is Some Text
# doing the same as
echo This Is Some Text | some-prog.js
```

Which is going to confuse a lot of people.

sindresorhus: > Usually programs that accept filenames as arguments are the ones that employ the use of stdin.

Sure, but I also often use it for CLI's that accept text input too, to make it easier for everyone to use it.

Then they could use either of these:

```
$ cli-app foo
$ echo foo | cli-app
```

index.js
```diff
@@ -8,11 +8,15 @@ var redent = require('redent');
...
 var loudRejection = require('loud-rejection');
 var normalizePackageData = require('normalize-package-data');
+var getStdin = require('get-stdin');
+var Promise = require('pinkie-promise');
 
 // get the uncached parent
 delete require.cache[__filename];
 var parentDir = path.dirname(module.parent.filename);
 
+global.Promise = Promise;
```
vdemedes: Isn't it against the rules or am I missing something?

index.js
```diff
@@ -80,6 +84,19 @@ module.exports = function (opts, minimistOpts) {
...
+			if (!_ && process.stdin.isTTY) {
+				return Promise.reject(new Error('input required'));
+			}
+
+			if (_.length > 0) {
+				return Promise.resolve(_);
+			}
+
+			return getStdin().then(function (data) {
+				return [data];
```
vdemedes: Why is array returned instead of data?

fixture.js
```diff
@@ -16,3 +16,14 @@ var cli = meow({
 Object.keys(cli.flags).forEach(function (el) {
 	console.log(el);
 });
+
+cli.stdinOrInput()
```
vdemedes: Maybe `cli.autoInput()` or `cli.getInput()` or something along those lines would look better and shorter?

index.js
SamVerschueren: Because `_` is an array as well. This way you can just iterate over the result instead of checking `Array.isArray`.

index.js
SamVerschueren: `get-stdin` does not support `0.10` and thus does not come bundled with `pinkie-promise`. So either we drop `0.10` in `meow`, or add `pinkie-promise` to `get-stdin` in order to get rid of this.

fixture.js
kevva: :+1: for `getInput()`.

index.js
```diff
@@ -80,6 +84,19 @@ module.exports = function (opts, minimistOpts) {
...
 		pkg: pkg,
 		help: help,
-		showHelp: showHelp
+		showHelp: showHelp,
+		stdinOrInput: function () {
+			if (!_ && process.stdin.isTTY) {
+				return Promise.reject(new Error('input required'));
+			}
+
+			if (_.length > 0) {
```
kevva: `_.length !== 0`

test.js
sindresorhus: Actually, I take back my last comment. Better to use stdin for just string input. No point in doing both, and stdin should be preferred as it works nicely in a pipe chain.

fixture.js
sindresorhus: 👍

PR #31, opened 2016-01-31
Allow running meow() more than once

rstacruz:
Useful for tests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,9 +14,13 @@ var normalizePackageData = require('normalize-package-data');
 // get the uncached parent
 delete require.cache[__filename];
 var parentDir = path.dirname(module.parent.filename);
+var loudRejectionEnabled;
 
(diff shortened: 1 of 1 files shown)


rstacruz:
Ahhh.. That's better.

sindresorhus:
`loud-rejection` already does this check, but I've removed the trace warning about it being called multiple times. I assume that will fix it for you.

rstacruz:
yes it does, thank you.

fwiw, my use case was trying to make my bins structured in this way:

```js
#!/usr/bin/node
/* bin/myapp.js */

function run (args) {
  meow({ ... })
}

module.exports = run
if (!module.parent) run(process.argv.slice(2))
```

...and then using it for tests as well.

sindresorhus:
Btw, you can `require()` bin files? Node.js strips the hashbang, but I guess your way is a bit cleaner.

Issue #33, opened 2016-02-16
Should not camelcase key '--'


egoist:
when passing minimistOpts `{'--': true}` and run a given example:

`cli whatever -- ava test.js`

`meow` returns:

```js
{
  ...
  flags: {
    '': ['ava', 'test.js']
  }
  // while it should be '--': ['ava', 'test.js']
}
```

PR #37, opened 2016-04-11
Skip camelcasing `--` option

kevva:
Since `minimist` accepts a `--` option we should probably not camelcase it since it'll remove it altogether as noted in #33.


sindresorhus:
Great! All that's left for the next major now is to resolve https://github.com/sindresorhus/meow/pull/22. Thoughts welcome there.

PR #43, opened 2016-05-13
Remove calls to process.exit() (fixes #42)

### LINKED ISSUE #42: help text is truncated after ~988 chars
I'm on Node 6. The output buffer isn't flushed on `process.exit()`.

It seems this problem was already encountered in `eslint` and fixed here: https://github.com/eslint/eslint/blob/master/bin/eslint.js#L80

I've applied a quick fix to my local copy and the issue goes away. The unit test for checking help output does not catch this issue after updating the fixture with very long help text.

rygine:
This should resolve the output buffering issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,6 +13,8 @@ var normalizePackageData = require('normalize-package-data');
 // get the uncached parent
 delete require.cache[__filename];
 var parentDir = path.dirname(module.parent.filename);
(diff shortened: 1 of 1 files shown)


rygine:
amended with style fix

rygine:
going to close this as the solution lies in the hands of the node team.

PR #52, opened 2016-10-26
Update dependencies

SimenB:
AFAICT all majors are just dropping support for node < 4, which this module has already done.

This change would make stuff dedupe better for us 😄 

Tests pass, but linting fails. I don't think that's my fault? `xo --version` gives `0.17.0`

### DIFF
--- a/index.js
+++ b/index.js
@@ -58,11 +58,13 @@ module.exports = (opts, minimistOpts) => {
 
 	const showHelp = code => {
 		console.log(help);
+		// eslint-disable-next-line unicorn/no-process-exit
 		process.exit(typeof code === 'number' ? code : 2);
 	};
 
 	if (argv.version && opts.version !== false) {
 		console.log(typeof opts.version === 'string' ? opts.version : pkg.version);
+		// eslint-disable-next-line unicorn/no-process-exit
 		process.exit();
 	}
 
--- a/package.json
+++ b/package.json
@@ -38,18 +38,18 @@
     "console"
   ],
   "dependencies": {
-    "camelcase-keys": "^3.0.0",
+    "camelcase-keys": "^4.0.0",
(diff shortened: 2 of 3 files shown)


SimenB:
Added a commit fixing two of the linting errors, but adding the shebang to fix the last one makes eslint import plugin barf 

```
 test.js:5:15
  ✖  5:15  Parse errors in imported module ./: Unexpected character # (1:1)  import/no-named-as-default
  ✖  5:15  Parse errors in imported module ./: Unexpected character # (1:1)  import/no-named-as-default-member
  ✖  5:15  Parse errors in imported module ./: Unexpected character # (1:1)  import/default
  ✖  5:15  Parse errors in imported module ./: Unexpected character # (1:1)  import/namespace
```

index.js
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env node
```
sindresorhus: That's not a correct fix. You can't just add this to a non-binary.

The correct fix is to add a inline `// eslint-disable-line unicorn/no-process-exit` comment.

package.json
```diff
@@ -38,18 +38,18 @@
...
-    "decamelize-keys": "^1.0.0",
-    "loud-rejection": "^1.0.0",
-    "minimist": "^1.1.3",
-    "normalize-package-data": "^2.3.4",
-    "read-pkg-up": "^1.0.1",
+    "camelcase-keys": "^4.0.0",
+    "decamelize-keys": "^1.1.0",
+    "loud-rejection": "^1.6.0",
+    "minimist": "^1.2.0",
+    "normalize-package-data": "^2.3.5",
```
sindresorhus: No need to bump patch and minor versions.

SimenB:
@sindresorhus Updated

PR #53, opened 2016-11-23
Parse Aliases out of the box

### LINKED ISSUE #50: Parse Aliases
I know that now to parse need use [aliases](https://github.com/sindresorhus/aliases) package (related #24).

One of my projects consists of scripts (about a hundred) to run in Continuous Integration.

I have to add the same lines in each file.

``` js
const aliases = require('aliases');

meow(help, { alias: aliases(help, {h: 'help'}) });
```

I think that we can add `parseAliases` option to simplify API.

``` js
const help = `
    Usage
      $ unicorn <name>

    Options
      -r, --rainbow  Lorem ipsum dolor sit amet
      -m, --magic    Aenean commodo ligula eget dolor
`;

meow(help, { parseAliases: true });
```

I'd really like to have this feature by default.

We can parse aliases if `alias` option is not specified. It won't break backward compatibility.

a-x-:
closes #50 Parse Aliases @blond

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ const redent = require('redent');
 const readPkgUp = require('read-pkg-up');
 const loudRejection = require('loud-rejection');
 const normalizePackageData = require('normalize-package-data');
+const aliases = require('aliases');
 
 // prevent caching of this module so module.parent is always accurate
 delete require.cache[__filename];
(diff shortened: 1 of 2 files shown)


a-x-:
not tested still

sindresorhus:
Needs tests and docs, and Travis is failing.

PR #55, opened 2016-12-06
chore: add typescript typings

KnisterPeter:
Create typings for meow. These depends on minimist, therefore add
minimist typings to package dependencies.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,26 @@
+/// <reference types="minimist" />
+import * as minimist from 'minimist';
+
+declare function meow(options: string | string[] | meow.Options, minimistOptions?: minimist.Opts): meow.Result;
+declare namespace meow {
+
+	export interface Options {
+		description?: string|boolean;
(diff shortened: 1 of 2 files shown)


SamVerschueren:
I'd suggest adding them to DefinitelyTyped instead. It's still up to @sindresorhus to decide if he wants this but I'm 99% sure he says no. The reason is quite simple. He doesn't care about TypeScript :p. Actually, I think the real reason is because module maintainers can't support everything. Today it's TypeScript and Flow, tomorrow it's something else. So just keep it open untill Sindre replies but we would be happy to 👍 your PR on DefinitelyTyped.

KnisterPeter:
I would do if he don't like to add it here. Just wanted to give it a shot before adding it at DT.

sindresorhus:
What Sam said, sorry. It's not that I don't care about TypeScript specifically. I'm just not ready for the churn of having multiple type definitions in all my modules and having to maintain those, when I don't even use either.

PR #57, opened 2017-02-17
Add `parseAliases` option.

### LINKED ISSUE #50: Parse Aliases

wtgtybhertgeghgtwtg:
Proposed to close #50.  Is the test sufficient?  Is the documentation okay?  Would you like me to submit a PR to bump the `node` version of `aliases` before this?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -134,6 +134,27 @@ Infer the argument type.
 
 By default, the argument `5` in `$ foo 5` becomes a string. Enabling this would infer it as a number.
 
+##### parseAliases
+
+Type: `boolean`<br>
+Default: `false`
+
+Parse aliases in `help`.
+
+```js
+const cli = meow({
+	help: `
+		Usage
+		  $ foo <input>
+
+		Options
+		  -a --b Aliased command
+	`,
+	// -a will be an alias for --b
+	parseAliases: true
+});
+```
+
 #### minimistOptions
 
 Type: `Object`<br>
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -134,6 +134,27 @@ Infer the argument type.
...
+	help: `
+		Usage
+		  $ foo <input>
+
+		Options
+		  -a --b Aliased command
+	`,
+	// -a will be an alias for --b
+	parseAliases: true
+});
```
kevva:
```js
const cli = meow(`
	Usage
	  $ foo <input>

	Options
	  -r, --rainbow Lorem ipsum dolor sit amet
`, {
	// -r will be an alias for --rainbow
	parseAliases: true
});
```

PR #61, opened 2017-05-09
Failing test marked + funwithflags

aretecode:
-  🚩 swap minimist with [funwithflags](https://github.com/fluents/funwithflags)
- 👕 fix comment casing for lint rules
- 🚨 mark failing test that sets on `process.title`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false
 language: node_js
 node_js:
+  - '7'
   - '6'
   - '4'
--- a/index.js
+++ b/index.js
@@ -1,6 +1,6 @@
 'use strict';
 const path = require('path');
-const minimist = require('minimist');
+const fwf = require('funwithflags');
 const camelcaseKeys = require('camelcase-keys');
 const decamelizeKeys = require('decamelize-keys');
 const trimNewlines = require('trim-newlines');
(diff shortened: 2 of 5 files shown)


sindresorhus:
I don't want to switch from `minimist`. `minimist` is very mature and stable and I don't really see any upsides to this. Thanks for the suggestion though.

aretecode:
@sindresorhus funwithflags is simply minimist updated with more tests added, converted to ava, more documentation, the api is 1:1 all tests are the same it is 125% faster.

aretecode:
@sindresorhus that failing test marked in this pr is unrelated to any dependency change though, it happens with a fresh clone on node 7.10 - any idea why it is failing?

sindresorhus:
I just don't see it being worth switching now. That's just churn for very little benefit. CLI flag parsing is not a bottleneck, so slightly better perf is not important.

> that failing test marked in this pr is unrelated to any dependency change though, it happens with a fresh clone on node 7.10 - any idea why it is failing?

No idea. I've skipped the test in master for now. I don't have time to look into it. Seems to only happen on Node.js 7. Works fine on Node.js 6 and 4.

aretecode:
@sindresorhus understandable, thought I'd see if it was worth doing.

Yes that's why I added the 7 entry in travis.

Thanks for checking it :-)

PR #63, opened 2017-05-31
Rewrite flags input

vadimdemedes:
Uses [minimist-options](https://github.com/vadimdemedes/minimist-options) to define minimist options for flags more comfortably.

Before:

```js
const cli = meow({
	description: 'Custom description',
}, {
	alias: {
		u: 'unicorn'
	},
	default: {
		meow: 'dog',
		camelCaseOption: 'foo'
	}
});
```

After:

```js
const cli = meow({
	description: 'Custom description',
	flags: {
		unicorn: {alias: 'u'},
		meow: {default: 'dog'},
		camelCaseOption: {default: 'foo'}
	}
});
```

Note to @sindresorhus: removed the docs from readme, because you wanted to bundle args documentation inside.

### DIFF
--- a/fixture.js
+++ b/fixture.js
@@ -7,14 +7,11 @@ const cli = meow({
 	help: `
 		Usage
 		  foo <input>
-	`
-}, {
-	alias: {
-		u: 'unicorn'
-	},
-	default: {
-		meow: 'dog',
-		camelCaseOption: 'foo'
+	`,
+	flags: {
+		unicorn: {alias: 'u'},
+		meow: {default: 'dog'},
+		camelCaseOption: {default: 'foo'}
 	}
 });
 
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 'use strict';
 const path = require('path');
+const buildMinimistOptions = require('minimist-options');
 const minimist = require('minimist');
 const camelcaseKeys = require('camelcase-keys');
 const decamelizeKeys = require('decamelize-keys');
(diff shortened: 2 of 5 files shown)


SamVerschueren:
Awesome! This looks much much better!

@sindresorhus Might be a good idea to update [aliases](https://github.com/sindresorhus/aliases) as well. Or maybe we should create a new one?

sindresorhus:
I'm also not sure about the special `arguments` name. It kinda feels weird to have it in the `flags` option. Maybe we should just have a separate top-level option for it here in Meow. So it would be:

```js
const cli = meow({
	arguments: 'string',
	flags: {
		foo: {
			type: 'boolean',
			alias: 'f'
		}
	}
});
```

Maybe also call it `input`, to match the `cli.input` property?

vadimdemedes:
Agree about `arguments` => `input`.

vadimdemedes:
Re issue with an alias, I couldn't reproduce it. I added and pushed a test to be sure and demo it. Plus, several other existing tests validate the alias functionality.

sindresorhus:
Weird. I'll try to create a failing test.

sindresorhus:
Ah, it's because I have the options as the second argument, so they're not read at all.

```js
const cli = meow(`
	Usage
	  $ internal-ip
`, {
	flags: {
		foo: {
			type: 'boolean',
			alias: 'f'
		}
	}
});
```

The above should be supported too.

vadimdemedes:
Added support for specifying options both as first and second arguments.

sindresorhus:
@vadimdemedes Sorry totally forgot about this. I'll write some docs for it tomorrow.

vadimdemedes:
@sindresorhus No probs ;) Glad it got merged in!

PR #68, opened 2017-09-06
Add `autoVersion` and `autoHelp` options

### LINKED ISSUE #62: Allow subcommands to handle --help
Currently when handling subcommands using separate meow instances like this:
```js
const bar = argv => {
	const cli = meow({
		argv,
		help: `
		  Usage
		    $ foo bar <input>
	`});
};

const cli = meow(`
	Usage
	  $ foo <command>

	Commands
	  bar
`);

// Check cli.input and handle commands
```
If the user enters ```$ foo bar --help``` they get the help for ```foo``` and not ```foo bar```

There's a wrapper called [moggie](https://github.com/florianb/moggie) that handles subcommands, but I was thinking we could to add a ```commands: [commands]``` option to meow so it doesn't block the --help flag?

Meow could either:
 - Not show help if a command is matched, leaving everything else to the user to handle
 - Handle the argv filtering and return the matched command and filtered argv

I'm happy to write something and submit a PR.

angeloashmore:
Fixes #62 

Adds `autoVersion` and `autoHelp` options. If either is set to false, meow won't handle printing on `--version` and `--help`. This would allow for explicit handling of showing either, possibly in a subcommand context.

**Note**: I had to add an extra `\n` to the `spawn cli and show help screen` test to pass. Not sure if this due to my shell?

### DIFF
--- a/fixture.js
+++ b/fixture.js
@@ -7,7 +7,9 @@ const cli = meow({
 	help: `
 		Usage
 		  foo <input>
-	`
+  `,
+  autoVersion: !process.argv.includes('--no-auto-version'),
+  autoHelp: !process.argv.includes('--no-auto-help')
 }, {
 	alias: {
 		u: 'unicorn'
--- a/fixture.js
+++ b/fixture.js
@@ -8,8 +8,8 @@ const cli = meow({
 		Usage
 		  foo <input>
   `,
-  autoVersion: !process.argv.includes('--no-auto-version'),
-  autoHelp: !process.argv.includes('--no-auto-help')
+  autoVersion: process.argv.indexOf('--no-auto-version') === -1,
+  autoHelp: process.argv.indexOf('--no-auto-help') === -1
 }, {
 	alias: {
 		u: 'unicorn'
--- a/fixture.js
+++ b/fixture.js
@@ -8,8 +8,8 @@ const cli = meow({
 		Usage
 		  foo <input>
   `,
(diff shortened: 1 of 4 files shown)


sindresorhus:
You can already do this with the `help` and `version` option.

angeloashmore:
By setting the values to false? I would still like meow to handle help and version messages, but not in every case. Sorry, maybe I'm missing something?

Here's my use case:

```
$ myapp --help
```

I would like this to print the help message defined for `myapp`.

```
$ myapp subcommand --help
```

`myapp` calls `myapp-subcommand` in a child process and passes the arguments. I would like `myapp-subcommand`'s help message to be printed, but `myapp` is seeing the `--help` and printing it's own message.

This PR would allow me to disable help and version handling in meow, but still use the `showHelp` function when necessary.

Another option is to alias the help and version flags in `myapp` to something else and handle the flags manually, but I thought this might be another option.

angeloashmore:
@sindresorhus any thoughts on this? I'll close it out if you feel it's unnecessary. 🙂

sindresorhus:
Sorry for the late response. Yes, this makes sense. Can you remove the `false` support in the `help` and `version` options, as it's no longer needed with this option.

angeloashmore:
Sorry, I'm way late on revisiting this. Just pushed some commits to the PR.

- Removes false support for help and version
- Adds `showVersion` function to complement `showHelp`
- Updates the documentation

sindresorhus:
Thanks :)

PR #73, opened 2017-11-21
Add a new option 'psTitle'

boxersb:
Add a new option called 'psTitle', which is involved in the process.title change.

- psTitle : true `default`
  - `process.title` is pkg.name or pkg.bin
- psTitle : false
  - `process.title` will be natural.
- psTitle : 'myApp' (custom string)
  - `process.title` is custom title from you assigned.

### DIFF
--- a/index.js
+++ b/index.js
@@ -30,7 +30,8 @@ module.exports = (helpMessage, opts) => {
 		argv: process.argv.slice(2),
 		inferType: false,
 		input: 'string',
-		help: helpMessage
+		help: helpMessage,
+		psTitle: true
 	}, opts);
 
 	let minimistOpts = Object.assign({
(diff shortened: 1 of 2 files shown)


sindresorhus:
Why is this option needed? What's you use-case?

boxersb:
When I type command like `ps -ef`, CMD section looks like this.
```bash
> ps -ef
UID         PID   PPID  C STIME TTY          TIME CMD
root          1      0  0 11월20 ?      00:00:00 npm
root         16      1  0 11월20 ?      00:00:07 sample-nodejs
root        203      0  0 11월20 ?      00:00:00 /bin/bash
root       1217      0  0 11:44 ?        00:00:00 /bin/bash
root       1385   1217  0 11:45 ?        00:00:00 ps -ef
```

`sample-nodejs` is node application, that name of package.json, which used `meow`.

I want to expose full command of application.

sindresorhus:
How is that incorrect? It shows the name of your project. I'm trying to understand why you want it to be different from the project name.

boxersb:
It's not incorrect, but in my case (or usual case), full command name in ps list needed.  

Actually, I'm running my app in docker container. so, I want to check out my app running command, not app name.   

It's just suggestion. If you can consider this feature, please merge this PR, or not, you can just close it. :)

sindresorhus:
I'm gonna pass on this as it's not a common use-case and you can just set `process.title` yourself after the `meow` call.

PR #75, opened 2018-01-28
Add test to check boolean flag type enforcement

### LINKED ISSUE #74: Type is not always enforced
From https://github.com/marionebl/svg-term-cli/commit/fe5d2a6073ea78f259bbc5ebb0a8296b0c6599bf#r26544549

Test case:

```
node test.js --cursor=false
```

```js
// test.js
const cli = require('meow')(``, {
    flags: {
        cursor: { type: 'boolean' }
    }
});
console.log(typeof cli.flags.cursor); // string
```

It should output `false`.

macklinu:
Resolves #74

I was unable to reproduce #74 with meow v4.0.0 - both the example in the issue and through this unit test - and thought submitting a PR with the test could be helpful. Let me know if there's a better place for this test in the codebase or if there's anything I can do here to help with #74. Thanks. 🙂

### DIFF
--- a/test.js
+++ b/test.js
@@ -106,6 +106,16 @@ test('type inference', t => {
 	}).input[0], 5);
 });
 
+test('enforces boolean flag type', t => {
+	const cli = m(``, {
+		argv: ['--cursor=false'],
+		flags: {
+			cursor: {type: 'boolean'}
+		}
+	});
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for looking into this and doing the test :)

PR #77, opened 2018-02-15
Do not define unset boolean args

pvdlg:
Proposal to solve xojs/xo#299.

For each boolean arg that doesn't define a `default`, set it to `undefined`. After parsing remove the options with a value of `undefined`.

Unfortunately this doesn't work with `minimist` due to an unrelated bug when using both `default` and `alias`:
```js
minimist(['-f'], {boolean: ['foo'], alias: {f: 'foo'}, default: {foo: null}});
// => {_: [], foo: true, f: [null, true]}
// Should be {_: [], foo: true, f:  true}

minimist(['--foo'], {boolean: ['foo'], alias: {f: 'foo'}, default: {foo: null}});
// => {_: [], foo: true, f: [null, true]}
// Should be {_: [], foo: true, f:  true}
```

I replaced `minimist` with [yargs-parser](https://github.com/yargs/yargs-parser) to solve this issue.

`meow` relies on [minimist-options](https://github.com/vadimdemedes/minimist-options) to transform the options format. It still works with `yargs-parser` as the format is the same.

This PR works as a fix for xojs/xo#299 but it might be a desireable to support the [other yargs-parser options](https://github.com/yargs/yargs-parser#requireyargs-parserargs-opts). That would require to support those option in `minimist-options` or to replace it.

Not sure if it worth the effort though, as [yargs](https://github.com/yargs/yargs) already allow all those options and support the same convenient format as `meow` via the [options function](http://yargs.js.org/docs/#api-optionskey-opt):
```js
const argv = require('yargs')
    .options({
      rainbow: {
	type: 'boolean',
	alias: 'r'
     }
   }).argv;
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 'use strict';
 const path = require('path');
 const buildMinimistOptions = require('minimist-options');
-const minimist = require('minimist');
+const yargs = require('yargs-parser');
 const camelcaseKeys = require('camelcase-keys');
 const decamelizeKeys = require('decamelize-keys');
 const trimNewlines = require('trim-newlines');
@@ -35,9 +35,20 @@ module.exports = (helpMessage, opts) => {
 		autoVersion: true
 	}, opts);
 
+	const minimistFlags = opts.flags ? Object.keys(opts.flags).reduce(
+		(flags, flag) => {
+			if (flags[flag].type === 'boolean' && typeof flags[flag].default === 'undefined') {
+				flags[flag].default = undefined;
+			}
+
+			return flags;
+		},
+		opts.flags
+	) : opts.flags;
+
 	let minimistOpts = Object.assign({
 		arguments: opts.input
-	}, opts.flags);
+	}, minimistFlags);
 
 	minimistOpts = decamelizeKeys(minimistOpts, '-', {exclude: ['stopEarly', '--']});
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
Does Yargs also default to `undefined` if `boolean` is the type and no `default` is specified?

pvdlg:
Yards default to `false` like `minimist`.
For some reasons `minimist` ignore `{default: undefined}` but works with `{default: null}`.

sindresorhus:
I'm still a little bit unsure whether `undefined` when no `default` is specified for `boolean` is the best default. It's quite convenient not having to specify `default: false` for all boolean flags. Maybe we should make this behavior opt-in with a Meow option?

pvdlg:
OK, we can have an option. We just need to define what's the default.
What is most the common case:
-  Wanting the unset `boolean` arg to not be included in the result options object?
- Or wanting the unset `boolean` args to be `false`?

sindresorhus:
> Or wanting the unset boolean args to be false?

This. I've only needed unset options in XO and one other CLI before.

sindresorhus:
Even when `boolean` defaults to `false`, it should still be possible to unset them individually, by setting `default: null` or `default: undefined`.

pvdlg:
Yep. So we can add the option `booleanDefault`, and whatever value set there will be used as default, unless a default is already used.
By default `booleanDefault` will be `false`.

sindresorhus:
> So we can add the option booleanDefault, and whatever value set there will be used as default, unless a default is already used.
By default booleanDefault will be false.

Yes! Much better than what I had in mind.

pvdlg:
I added the `booleanDefault` option. By default it's `false` so the default behavior doesn't change.

pvdlg:
Updated to to [yargs-parser@10.0.0](https://github.com/yargs/yargs-parser/releases/tag/v10.0.0)

sindresorhus:
Looks good. Thanks @pvdlg :)

PR #85, opened 2018-04-25
Solves TypeError when using meow with zeit/pkg

### LINKED ISSUE #84: Error using meow with zeit/pkg
When using meow with [zeit/pkg](https://github.com/zeit/pkg), I'm receiving the error:
```
TypeError: Cannot read property 'scripts' of undefined
    at normalize (C:\snapshot\...\node_modules\normalize-package-data\lib\normalize.js:25:12)
    at module.exports (C:\snapshot\...\node_modules\meow\index.js:55:2)
```
This can be solved by replacing
```
const pkg = opts.pkg;
```
with
```
const pkg = opts.pkg || {};
```
at the [line 50](https://github.com/sindresorhus/meow/blob/8c82a18c70e1081771d97f88a6d5f5ce9b00005d/index.js#L50).

thiagodp:
Closes #84

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,7 +47,7 @@ module.exports = (helpMessage, opts) => {
 
 	minimistOpts = buildMinimistOptions(minimistOpts);
 
-	const pkg = opts.pkg;
+	const pkg = opts.pkg || {};
 	const argv = minimist(opts.argv, minimistOpts);
 	let help = redent(trimNewlines((opts.help || '').replace(/\t+\n*$/, '')), 2);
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -47,7 +47,7 @@ module.exports = (helpMessage, opts) => {
 
 	minimistOpts = buildMinimistOptions(minimistOpts);
 
-	const pkg = opts.pkg;
+	const pkg = opts.pkg || {};
```
sindresorhus: Better to place this on line 29, where the other logic is.

thiagodp:
Updated

PR #88, opened 2018-05-23
Add badge to display install size

styfle:
This adds a badge to the `README.md` that displays the npm install size


sindresorhus:
Thanks, but I'm not interested in having this badge on all my repos. Meow is for CLI tools where size doesn't matter as much. I accepted it on Ow because it can also be used in the browser.

I would recommend you instead create a browser extension that injects the npm package install size info into every npm package repo on GitHub. That would be useful.

styfle:
My [long term goals](https://github.com/styfle/packagephobia#what-are-the-long-term-goals) are to get this info on npmjs.com or npms.io so that it's right next to all the other stats like downloads, maintenance, etc.

Package Phobia is actually [meant](https://github.com/styfle/packagephobia#demo) for CLI tools, not browser based tools. This is because it's measure the entire npm install size, not the bundle size you ship to the browser.

That's why you can find this badge in the readme for [webpack](https://github.com/webpack/webpack/blob/master/README.md), [rollup](https://github.com/rollup/rollup#readme), and many others.

PR #91, opened 2018-07-03
Fix documentation regarding aliases

Zhouzi:
Correct me if I am wrong but I believe that the current documentation regarding aliases is incorrect.


sindresorhus:
No, the `alias` is part of the flags object: https://github.com/sindresorhus/meow/blob/258659a6e4cf102ef09a7c27efcee1e953808725/test.js#L169 What makes you think otherwise?

Zhouzi:
Okay my bad, the problem must come from the code surrounding the project I'm working on 👍

PR #100, opened 2018-09-27
Add rollup build to create a package without dependencies

### LINKED ISSUE #67: Bundle dependencies
I usually don't care to do this, but it would be nice if `meow` was simply one dependency as it would help the install time on lots of small CLIs.

Not exactly sure the best way to do this. It should be something that doesn't add any build-script for development, meaning, everything should be handled in a `prepublish` script.

Maybe [`npm-packager`](https://github.com/sheerun/npm-packer) or just build it with Webpack.

Suggestions and help welcome.

pimlie:
This resolves #67 

It also makes the installed size a bit smaller (at least on my filesystem). On my fs with a 4KB block size it goes from `~1.3MB` to `~200KB`.


package.json
```diff
@@ -37,24 +42,35 @@
...
 		"yargs-parser": "^10.0.0"
 	},
-	"devDependencies": {
-		"ava": "*",
-		"execa": "^0.10.0",
-		"indent-string": "^3.0.0",
-		"xo": "*"
-	},
 	"xo": {
+		"semicolon": false,
```
sindresorhus: Don't do unrelated changes. Same applies to all the removing of semicolons. You might not like them, but you should follow what the project already uses.

package.json
```diff
@@ -13,10 +14,14 @@
 		"node": ">=6"
 	},
 	"scripts": {
-		"test": "xo && ava"
+		"build": "rollup -c build/rollup.config.js",
```
sindresorhus:
```suggestion
		"build": "rollup --config=build/rollup.config.js",
```

build/rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+// -----------------------------
+// Aliases
+// -----------------------------
+const rootDir = resolve(__dirname, '..');
+
+export default {
+	input: resolve(rootDir, 'lib', 'index.js'),
+	output: {
+		name: 'meow',
+		file: resolve(rootDir, 'dist', 'meow.js'),
```
sindresorhus: The file should still be named `index.js`.

build/rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+const version = packageJson.version || process.env.VERSION;
+
+// -----------------------------
+// Banner
+// -----------------------------
+const banner =
+  '/*!\n' +
+  ' * meow v' + version + '\n' +
+  ' * Released under the MIT License.\n' +
+  ' */';
```
sindresorhus: Don't need a banner. The license file is included in the package.

package.json
```diff
@@ -13,10 +14,14 @@
...
+		"lint": "xo lib test",
+		"test": "yarn build && yarn lint && yarn test:lib && yarn test:dist",
+		"test:lib": "ava test/lib.test.js",
+		"test:dist": "ava test/dist.test.js"
```
sindresorhus: The dist file should be built at the `prepublish` step. The dist files should not be committed.

package.json
```diff
@@ -13,10 +14,14 @@
 		"node": ">=6"
 	},
 	"scripts": {
-		"test": "xo && ava"
+		"build": "rollup -c build/rollup.config.js",
+		"lint": "xo lib test",
+		"test": "yarn build && yarn lint && yarn test:lib && yarn test:dist",
```
sindresorhus: Don't compile the test files. They can just be source. Only the `index.js` should be rolled up.

sindresorhus:
Sorry about the slow reply. Thanks for working on this :)

package.json
pimlie: The dist files are not commited afaik as `dist` is in .gitignore?

package.json
pimlie: I am not compiling the test files afaik (I only lint them)

pimlie:
@sindresorhus Today I had finally some time to look at this again, so now I am sorry for the delay.

I have implemented some changes, including a better implementation for testing both lib as dist as they will now both always use the same tests. I am not sure how you feel about testing both, but I always feel more confident to also test the dist package as e.g. a wrong rollup config could result in a broken dist. But maybe that requires some solutions which are to creative for your taste ;)

Let me know if you have any follow-up remarks!

pimlie:
As this issue has become quite stale I am closing it

Issue #102, opened 2018-11-05
Remove short-flags in `cli.flags` if they're defined as `alias`


sindresorhus:
```js
const cli = m({
	argv: ['-c', '-l'],
	flags: {
		coco: {
			alias: 'c'
		},
		loco: {
			alias: 'l'
		}
	}
});

console.log(cli.flags);
//=> { c: true, coco: true, l: true, loco: true }
```

Since `c` and `l` are defined as `alias` for two other flags, I think we should just exclude the short-flags from the `cli.flags` output. They just create noise there.

So it would be this instead:

```js
console.log(cli.flags);
//=> { coco: true, loco: true }
```

Thoughts?

IssuehuntBot:
@issuehuntfest has funded $40.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/25121561/issues/102)

IssuehuntBot:
@picchikevin has submitted a pull request. [See it on IssueHunt](https://issuehunt.io/repos/25121561/issues/102)

IssuehuntBot:
@sindresorhus has rewarded $36.00 to @picchikevin. [See it on IssueHunt](https://issuehunt.io/repos/25121561/issues/102)

- :moneybag: Total deposit: $40.00
- :tada: Repository reward(0%): $0.00
- :wrench: Service fee(10%): $4.00

revelt:
hi all, around 10 meow dependent CLI's of mine were dependent on aliases. I've been using `if (cli.flags.z)` all over the code for convenience. What was done was done, v6 is out and I had to fix bunch of CLI's but in the future **please don't remove issues if they already exist and don't block anything**.

sindresorhus:
> I've been using if (cli.flags.z) all over the code for convenience.

Convenience for you, but worse readability for anyone else reading your code. The change was done exactly to fix code like yours.

> please don't remove issues if they already exist and don't block anything.

I cannot promise that. If something will help people improve their CLIs or code, I'll do it.

PR #108, opened 2019-01-10
Remove flags aliases from flags output

PicchiKevin:
As said in this issue #102, short-flags in the `cli.flags` are a bit messy

### DIFF
--- a/index.js
+++ b/index.js
@@ -109,6 +109,12 @@ module.exports = (helpText, options) => {
 
 	const flags = camelcaseKeys(argv, {exclude: ['--', /^\w$/]});
 
+	Object.keys(options.flags).forEach(function(flag){
+		if(options.flags[flag].alias != undefined){
+			delete flags[options.flags[flag].alias];
+		}
+	});
+
 	return {
 		input,
 		flags,


PicchiKevin:
This PR add `cli.simpleFlags`, the only difference with `cli.flags` is that it wont output flags aliases.

Example:

```Javascript
const cli = m({
	argv: ['-c', '-l'],
	flags: {
		coco: {
			alias: 'c'
		},
		loco: {
			alias: 'l'
		}
	}
});

console.log(cli.flags);
//=> { c: true, coco: true, l: true, loco: true }

console.log(cli.simpleFlags);
//=> { c: true, coco: true}
```

index.js
```diff
@@ -108,10 +108,20 @@ module.exports = (helpText, options) => {
 	delete argv._;
 
 	const flags = camelcaseKeys(argv, {exclude: ['--', /^\w$/]});
+	const simpleFlags = camelcaseKeys(argv, {exclude: ['--', /^\w$/]});
```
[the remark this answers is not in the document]
sindresorhus: Instead, maybe we could do the opposite and add an new object for the full list of flags. Naming suggestions welcome. Something like `rawFlags` (Not really raw though as they're still camelcased), `unnormalizedFlags`,... ?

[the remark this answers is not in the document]
PicchiKevin: I did that to prevent a breaking changes.(My bad not seeing the 6.0.0 Milestone)
`unnormalizedFlags` is a good one (I'm open for any proposal)

index.js
```diff
@@ -108,10 +108,18 @@ module.exports = (helpText, options) => {
 	delete argv._;
 
 	const flags = camelcaseKeys(argv, {exclude: ['--', /^\w$/]});
+	const unnormalizedFlags = camelcaseKeys(argv, {exclude: ['--', /^\w$/]});
```
sindresorhus: It's wasteful to run the exact same function twice. You can just copy the `flags` object using object spread.

index.js
```diff
@@ -108,10 +108,18 @@ module.exports = (helpText, options) => {
...
+	if (options.flags !== undefined) {
+		for (const flag of Object.entries(options.flags)) {
+			delete flags[flag[1].alias];
+		}
+	}
 
 	return {
 		input,
 		flags,
+		unnormalizedFlags,
```
sindresorhus: I'm still not sold on this name. I think we should try to come up with something shorter.

index.js
```diff
@@ -108,10 +108,18 @@ module.exports = (helpText, options) => {
 	delete argv._;
 
 	const flags = camelcaseKeys(argv, {exclude: ['--', /^\w$/]});
+	const unnormalizedFlags = camelcaseKeys(argv, {exclude: ['--', /^\w$/]});
+
+	if (options.flags !== undefined) {
+		for (const flag of Object.entries(options.flags)) {
```
sindresorhus: `for (const flag of Object.entries(options.flags)) {` => `for (const flagValue of Object.values(options.flags)) {`

sindresorhus:
This also needs docs updates.

index.js
PicchiKevin: Didn't know that existed, thanks for teaching me something.
Could we do it this way ? 
```Javascript
	const flags = camelcaseKeys(argv, {exclude: ['--', /^\w$/]});
	const unnormalizedFlags = {...flags};
```

index.js
PicchiKevin: I've got `unsortedFlags`, what do you thing about it  🤷‍♂️ ?

PicchiKevin:
Made the changes requested but still can't choose a name.
I've got `unsortedFlags` 🤷‍♂️ 

Also updated in the doc, should I add more infos ?

index.js
sindresorhus: Not really better. Let's just go with `unnormalizedFlags` for now.

sindresorhus:
This looks good now. Thank you :)

PR #112, opened 2019-02-20
Added support for multiple option

### LINKED ISSUE #111: Add setting to indicate a flag can be set multiple times

Currently, if you have:

```js
foo: {
	type: 'string'
}
```

```
$ cli --foo=a
```

```
$ cli --foo=a --foo=b
```

```js
{
	flags: {
		foo: {
			type: 'string',
			multiple: true
		}
	}
}
```

b4dnewz:
When the `multiple` option is set **true** on string type flags the parsed value is always an array, closes #111 

#### Observations

I see you are using yargs-parsers but creating the options for it with minimist-options, I see they have pretty much the same config and it's working pretty good but I've noticed this difference while making this PR: 

The [array](https://github.com/vadimdemedes/minimist-options#usage) type of minimist-options turns into: `array: ['arr']` while the array options of yargs-parser is more complex and can take [objects](https://github.com/yargs/yargs-parser#api) for better definition of the expected arguments, so maybe this could lead to having the  `multiple` on boolean flags too.

Do you think is necessary to have minimist-options to parse flags for you or it can be substituted with a custom function that will complain more with yargs-parser style of options?

Sorry for my bad english :(

### DIFF
--- a/test.js
+++ b/test.js
@@ -81,6 +81,28 @@ test('single character flag casing should be preserved', t => {
 	t.deepEqual(meow({argv: ['-F']}).flags, {F: true});
 });
 
+test('string flags with multiple options are always arrays', t => {
+	const flags = {
+		foo: {
+			type: 'string',
+			multiple: true
+		}
+	};
+
+	t.deepEqual(meow({
+		flags,
+		argv: ['--foo=a']
+	}).flags, {foo: ['a']});
+
+	t.deepEqual(meow({
+		flags,
+		argv: [
+			'--foo=a',
+			'--foo=b'
+		]
+	}).flags, {foo: ['a', 'b']});
+});
+
 test('type inference', t => {
 	t.is(meow({argv: ['5']}).input[0], '5');
 	t.is(meow({argv: ['5']}, {input: 'string'}).input[0], '5');
--- a/test.js
+++ b/test.js
@@ -107,6 +107,11 @@ test('string flags with multiple options are always arrays', t => {
 
 	t.deepEqual(meow({
 		flags,
+		argv: ['--foo', 'a', 'b']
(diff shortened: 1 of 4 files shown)


b4dnewz:
I'm sorry about that.. I don't know while my editor eslint-plugin didn't take care of that errors
I've force-pushed to avoid having useless commit such "Fix lint errors" on the history 😄

medusalix:
> Instead, we should **allow only one value by default**, and add a setting to the flags config to allow multiple for a flag, in which case, it will always return an array, even if there's only one flag.

Your PR should probably handle that case too 😃. Only when `multiple` is specified, the flag should accept more than one value.

b4dnewz:
You right, fortunately there was [this option](https://github.com/yargs/yargs-parser#duplicate-arguments-array) on yars-parser that did the job.

mischah:
Thanks for working on this. 

One question: Will `multiple: true` also make ist possible to parse something like:

```
command --flag hu ha 
```

as array:

```
{ flag: ['hu', 'ha'] }
```

Pretty much like https://github.com/yargs/yargs/blob/master/docs/api.md#arraykey

test.js
```diff
@@ -105,6 +105,11 @@ test('string flags with multiple options are always arrays', t => {
 		argv: ['--foo=a']
 	}).flags, {foo: ['a']});
 
+	t.deepEqual(meow({
+		flags,
+		argv: ['--foo', 'a', 'b']
+	}).flags, {foo: ['a', 'b']});
```
mischah: ❤️

b4dnewz:
I made my own helper module https://github.com/b4dnewz/clito so I close this since I don't need it anymore

PR #114, opened 2019-02-22
Only consider enabling autoHelp/autoVersion in case there is only one argument in `process.argv`

### LINKED ISSUE #113: Only show version/help when `cli.input.length === 0` is true
Hi, it's a feature request!

I'm working on this issue: https://github.com/sindresorhus/latest-version-cli/issues/2.

About this case:

```bash
$ latest-version npm --version=beta
```

I think `meow` should not print the version if `cli.input.length === 0` is `false`.

Although we already have the `autoVersion` option, but if we turn it to `false` we have to handle the `showVersion` manually.

LitoMore:
Resolved #113

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -10,8 +10,8 @@
 - Parses arguments
 - Converts flags to [camelCase](https://github.com/sindresorhus/camelcase)
 - Negates flags when using the `--no-` prefix
-- Outputs version when only the `--version`
-- Outputs description and supplied help text when only the `--help`
+- Outputs version when `--version`
+- Outputs description and supplied help text when `--help`
 - Makes unhandled rejected promises [fail hard](https://github.com/sindresorhus/hard-rejection) instead of the default silent fail
 - Sets the process title to the binary name defined in package.json
 
@@ -260,6 +260,12 @@ Meow will make unhandled rejected promises [fail hard](https://github.com/sindre
 
 ## Tips
 
+### `autoVersion` and `autoHelp`
+
+The `autoVersion` and `autoHelp` option will only enabled when there is only one argument in argv.
+
+## Related
+
 See [`chalk`](https://github.com/chalk/chalk) if you want to colorize the terminal output.
 
 See [`get-stdin`](https://github.com/sindresorhus/get-stdin) if you want to accept input from stdin.
@@ -269,7 +275,6 @@ See [`conf`](https://github.com/sindresorhus/conf) if you need to persist some d
 See [`update-notifier`](https://github.com/yeoman/update-notifier) if you want update notifications.
 
 [More useful CLI utilities…](https://github.com/sindresorhus/awesome-nodejs#command-line-utilities)
-
 
 ## License
 
--- a/readme.md
+++ b/readme.md
@@ -262,7 +262,7 @@ Meow will make unhandled rejected promises [fail hard](https://github.com/sindre
 
 ### `autoVersion` and `autoHelp`
 
-The `autoVersion` and `autoHelp` option will only enabled when there is only one argument in argv.
+The `autoVersion` and `autoHelp` option will only enabled when there is only one argument in `process.argv`.
 
 ## Related
 
--- a/readme.md
+++ b/readme.md
@@ -10,8 +10,8 @@
 - Parses arguments
 - Converts flags to [camelCase](https://github.com/sindresorhus/camelcase)
 - Negates flags when using the `--no-` prefix
-- Outputs version when `--version`
-- Outputs description and supplied help text when `--help`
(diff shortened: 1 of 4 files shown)


sindresorhus:
The behavior needs to be documented.

test.js
```diff
@@ -45,6 +45,16 @@ test('spawn cli and not show version', async t => {
 	t.is(stdout, 'version\nautoVersion\nmeow\ncamelCaseOption');
 });
 
+test('spwan cli test input without version output', async t => {
```
sindresorhus: Typo

test.js
```diff
@@ -45,6 +45,16 @@ test('spawn cli and not show version', async t => {
...
+	const {stdout} = await execa('./fixture.js', ['bar', '--version']);
```
sindresorhus: You don't need to spawn for this kind of test. Better to just use the `meow` `argv` option. See other tests.

LitoMore:
@sindresorhus Code updated. I documented this behavior in the `Features` section, is that OK?

readme.md
```diff
@@ -10,8 +10,8 @@
 - Parses arguments
 - Converts flags to [camelCase](https://github.com/sindresorhus/camelcase)
 - Negates flags when using the `--no-` prefix
-- Outputs version when `--version`
-- Outputs description and supplied help text when `--help`
+- Outputs version when only the `--version`
```
sindresorhus: This doesn't sound good when read out loud. It's also not correct as you can have other flags together with `--version`, just not other arguments. Should we ensure there are no other flags too? Or is that too limiting? Any downsides doing it?

LitoMore: I imagined a case like this (it's just an example, not the real API):

```bash
# This feature is that we could only type `package-json` under a package,
# it will read package.json file automatically then return the result.

~/Git/npm foo-branch ⇡
❯ package-json --version=next --full-metadata --all-versions

{
	"name": "npm",
	"version": "6.9.0-next.0"
	...
}
```

In this case, we don't have any `cli.input` but have three flags for the command.

I think we should ensure there are no other flags and inputs for the version output. It should not be output even `version` has a value.

LitoMore: There are too many logics to explain for this issue. We could add a new section for document the issue, or document in the `autoVersion` and `autoHelp` section.

sindresorhus: > I think we should ensure there are no other flags and inputs for the version output. It should not be output even `version` has a value.

:+1:

> We could add a new section for document the issue

👍

LitoMore:
For the document, I renamed the `Tips` section to `Related`, and add a new section named `Tips` for `autoVersion` and `autoHelp` tips.

LitoMore:
@sindresorhus ping

LitoMore:
Any progress? @sindresorhus

sindresorhus:
> add a new section named Tips for autoVersion and autoHelp tips.

This should be document under the docs for those options, not in a Tips section. It's not a tip, it's documenting how something works.

readme.md
```diff
@@ -260,6 +260,12 @@ Meow will make unhandled rejected promises [fail hard](https://github.com/sindre
 
 ## Tips
 
+### `autoVersion` and `autoHelp`
+
+The `autoVersion` and `autoHelp` option will only enabled when there is only one argument in argv.
```
sindresorhus:
```suggestion
The `autoVersion` and `autoHelp` option will only enabled when there is only one argument in `process.argv`.
```

Would also be helpful to include why this is the case.

LitoMore:
@sindresorhus Docs updated

readme.md
```diff
@@ -148,13 +148,17 @@ Default: `true`
 
 Automatically show the help text when the `--help` flag is present. Useful to set this value to `false` when a CLI manages child CLIs with their own help text.
 
+This option will only enabled when there is only one argument in `process.argv`.
```
sindresorhus: This needs to be added to https://github.com/sindresorhus/meow/blob/master/index.d.ts too

sindresorhus:
```suggestion
This option is only considered when there is only one argument in `process.argv`.
```

readme.md
```diff
@@ -272,7 +276,6 @@ See [`update-notifier`](https://github.com/yeoman/update-notifier) if you want u
 
 [More useful CLI utilities…](https://github.com/sindresorhus/awesome-nodejs#command-line-utilities)
 
-
```
sindresorhus: Don't do unrelated changes.

sindresorhus:
The PR title needs to be updated to reflect the final behavior.

LitoMore:
@sindresorhus Updated

sindresorhus:
Thanks :)

PR #115, opened 2019-02-24
Using process.mainModule instead of module.parent to get module directory

### LINKED ISSUE #90: Throws when used with --experimental-modules
If using an mjs module using --experimental-modules and importing `meow` it throws on this line:

https://github.com/sindresorhus/meow/blob/258659a6e4cf102ef09a7c27efcee1e953808725/index.js#L15

I think this might be the Node issue: https://github.com/nodejs/node/issues/15760

ianwalter:
Sorry if you've already considered this but have decided against it for some reason. I couldn't find a discussion about it. I think this might be a cleaner solution and might also fix #90.

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,10 +10,6 @@ const readPkgUp = require('read-pkg-up');
 const hardRejection = require('hard-rejection');
 const normalizePackageData = require('normalize-package-data');
 
-// Prevent caching of this module so module.parent is always accurate
-delete require.cache[__filename];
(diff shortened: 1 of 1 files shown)


sindresorhus:
This could work in theory, but it makes it impossible to programmatically import CLIs, which can be useful for testing, as when `meow` is imported, it's no longer the `mainModule`.

ianwalter:
Ah, that's fair. I didn't think of that since I use execa to test CLIs. Closing.

PR #116, opened 2019-03-08
Add TypeScript definition

BendingBender:


index.d.ts
```diff
@@ -0,0 +1,195 @@
...
+
+	/**
+	 * Flags converted camelCase including aliases.
+	 */
+	unnormalizedFlags: {[name: string]: unknown};
+
+	/**
+	 * The `package.json` object.
+	 */
+	pkg: {[key: string]: unknown};
```
sindresorhus: Would be really cool if we could properly type the package.json.

https://docs.npmjs.com/files/package.json
http://json.schemastore.org/package

You could submit the definition to https://github.com/sindresorhus/type-fest so we could reuse it in other packages too.

What do you think?

---

We could also expose a type that extends the spec compliant one with a realistic definition that also supports keys like `types` and `eslintConfig`.

index.d.ts
```diff
@@ -0,0 +1,195 @@
...
+	 * 	}
+	 * }
+	 */
+	readonly flags?: MinimistOptions;
+
+	/**
+	 * Description to show above the help text. Default: The package.json `"description"` property.
+	 *
+	 * Set it to `false` to disable it altogether.
+	 */
```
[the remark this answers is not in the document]
bfred-it: I think they are the JSDoc style. Would this break VSCode's ability to display the description inline, for example?

> hard to modify the comments

That just depends on the editor and its plugins. I had no problem with Sublime Text (but I think I'm missing some plugin in VSCode)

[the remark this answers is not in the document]
BendingBender: This, however, disables markdown syntax highlighting in doc blocks in VSCode.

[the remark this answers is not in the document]
BendingBender: Should I drop the block-level indentation as well?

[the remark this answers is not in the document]
sindresorhus: > This, however, disables markdown syntax highlighting in doc blocks in VSCode.

I'm ok with that. The markdown is still correctly rendered in the preview for consumers.

https://microsoft.github.io/tsdoc/ correctly syntax highlights Markdown, even without the `*`, so it's probably just an oversight in VS Code.

index.d.ts
BendingBender: Sounds reasonable. Will do.

index.d.ts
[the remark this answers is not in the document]
sindresorhus: > Should I drop the block-level indentation as well?

The indentation should be the same as the code below.

BendingBender:
Waiting for https://github.com/sindresorhus/type-fest/pull/5.

sindresorhus:
> Waiting for sindresorhus/type-fest#5.

Published :)

index.d.ts
```diff
@@ -0,0 +1,197 @@
...
+
+	@example
+	```
+flags: {
+	unicorn: {
+		type: 'string',
+		alias: 'u',
+		default: 'rainbow'
+	}
+}
```
sindresorhus: Sorry for being unclear, this should be indented the same as the backticks.

```
	```
	flags: {
		unicorn: {
			type: 'string',
			alias: 'u',
			default: 'rainbow'
		}
	}
```

index.d.ts
```diff
@@ -0,0 +1,197 @@
...
+
+	_You most likely don't need this option._
+	 */
+	readonly pkg?: {[key: string]: unknown};
+
+	/**
+	Custom arguments object.
+
+	@default process.argv.slice(2)
+	 */
```
sindresorhus:
```suggestion
	*/
```

index.d.ts
```diff
@@ -0,0 +1,197 @@
...
+//		r: true,
+//		unicorn: false,
+//		u: false,
+//		sparkles: true
+//	},
+//	…
+//}
+	 ```
+	 */
+	readonly booleanDefault?: boolean | null;
```
sindresorhus: According to the docs this should support `undefined` too. It's unclear to me why `null` is mentioned in the docs though. Might be a mistake...

index.d.ts
```diff
@@ -0,0 +1,197 @@
...
+export interface Result {
+	/**
+	Non-flag arguments.
+	 */
+	input: string[];
+
+	/**
+	Flags converted to camelCase excluding aliases.
+	 */
+	flags: {[name: string]: unknown};
```
sindresorhus: What your opinion on `{[name: string]: unknown};` vs `Record<string, unknown>`?

index.d.ts
```diff
@@ -0,0 +1,197 @@
...
+	The help text used with `--help`.
+	 */
+	help: string;
+
+	/**
+	Show the help text and exit with code.
+
+	@param code The exit code to use. Default: `2`.
+	 */
+	showHelp(code?: number): void;
```
sindresorhus: `code` => `exitCode` would be clear (I know you just copied from the readme)

index.d.ts
BendingBender: It's already optional, so `undefined` is already included.

index.d.ts
BendingBender: I think `{[name: string]: unknown};` is more common, so I'd prefer this.

index.d.ts
sindresorhus: Ah right... My mistake.

index.d.ts
BendingBender: And it allows naming the key.

index.d.ts
BendingBender: Included it for clarity.

sindresorhus:
This looks great! :)

sindresorhus:
Just FYI. It's probably going to take some time until I can publish this one as it's blocked by doing the next major release.

PR #122, opened 2019-08-09
Add type information for flags

NiGhTTraX:
This PR adds some type safety around `flags` and `unnormalizedFlags`.

```ts
meow({ flags: { foo: 'number' } }).flags.foo // number
meow({ flags: { foo: { type: 'string' } } }).flags.foo // string
meow({ flags: { 'foo-bar': 'boolean' } }).unnormalizedFlags['foo-bar'] // boolean
```

The following use cases remain untyped due to limitations in the current TS version (can't manipulate strings):

```ts
meow({ flags: { 'foo-bar': 'number' } }).flags.fooBar // unknown
meow({ flags: { foo: { type: 'string', alias: 'f' } }).unnormalizedFlags.f // unknown
```

The following use case gives the wrong type info (again because we can't manipulate strings):

```ts
meow({ flags: { 'foo-bar': 'number' } }).flags['foo-bar'] // number, but actually undefined
```

@sindresorhus do you think it's worth adding types for the rest of the use cases, or will the fragmented (and potentially wrong) type info confuse users?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -160,11 +160,11 @@ declare namespace meow {
 	}
 
 	type TypedFlags<Flags extends MinimistOptions> = {
-		[F in keyof Flags]: Flags[F] extends { type: 'number' }
+		[F in keyof Flags]: Flags[F] extends {type: 'number'}
 			? number
-			: Flags[F] extends { type: 'string' }
+			: Flags[F] extends {type: 'string'}
 				? string
-				: Flags[F] extends { type: 'boolean' }
+				: Flags[F] extends {type: 'boolean'}
 					? boolean
 					: unknown;
 	};
--- a/index.d.ts
+++ b/index.d.ts
@@ -178,12 +178,12 @@ declare namespace meow {
 		/**
 		Flags converted to camelCase excluding aliases.
 		*/
-		flags: TypedFlags<Flags>;
+		flags: TypedFlags<Flags> & {[name: string]: unknown};
 
 		/**
 		Flags converted camelCase including aliases.
 		*/
-		unnormalizedFlags: {[name: string]: unknown};
+		unnormalizedFlags: TypedFlags<Flags> & {[name: string]: unknown};
 
 		/**
 		The `package.json` object.
--- a/index.d.ts
+++ b/index.d.ts
@@ -160,11 +160,11 @@ declare namespace meow {
 	}
 
 	type TypedFlags<Flags extends MinimistOptions> = {
-		[F in keyof Flags]: Flags[F] extends { type: 'number' } | 'number'
+		[F in keyof Flags]: Flags[F] extends { type: 'number' }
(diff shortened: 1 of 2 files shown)


sindresorhus:
I like this :)

---

> The following use case gives the wrong type info:

I think the benefits of strongly-typed properties outweighs this obscure edge-case. I'm fine with it.

> do you think it's worth adding types for the rest of the use cases

What else did you have in mind?

index.d.ts
```diff
@@ -159,7 +159,17 @@ declare namespace meow {
 		readonly hardRejection?: boolean;
 	}
 
-	interface Result {
+	type TypedFlags<Flags extends MinimistOptions> = {
+		[F in keyof Flags]: Flags[F] extends { type: 'number' } | 'number'
```
sindresorhus: Why the ` | 'number'` part? `flags` has to be an object.

NiGhTTraX: When I created this you could also define flags like `flags: { foo: 'number' }`. So here I'm checking if a flag is the string `'number'` or an object with `type: 'number'`. Did that change?

NiGhTTraX:
> > do you think it's worth adding types for the rest of the use cases

> What else did you have in mind?

I realize now that I phrased the question wrong. I meant to ask if adding types just for the first block of cases in the description is worth it.

index.d.ts
sindresorhus: Yeah. That's not supported anymore.

sindresorhus:
> I meant to ask if adding types just for the first block of cases in the description is worth it.

Yes, definitely.

index.d.ts
NiGhTTraX: OK. I've updated the code to reflect that, but `MinimistOptions` still allows defining `{flags: {foo: 'number'}}`. The possibility comes from

```ts
type Options<ArrayOptionContentType = unknown> = {
  [key: string]: OptionType /* | <other object types> */
type OptionType = 'string' | 'boolean' | 'number' | 'array';
```

Passing in `foo: 'number'` still works because `meow` is just calling `yargs`. Should the types be stricter to only allow the documented CLI? Reusing `MinimistOptions` is easy (and correct with the current codebase), but it doesn't align with the docs.

In any case, I don't think this is blocking this PR so we could move the discussion to a new issue, if it's even worth continuing the discussion.

sindresorhus: > Should the types be stricter to only allow the documented CLI?

Yes, ideally we should that kind of usage in the JS code too.

NiGhTTraX: > ideally we should that kind of usage in the JS code too.

I'm guessing you meant forbid here since the JS code already allows patterns that are not part of the docs. If I have time maybe I'll make a PR to first restrict the types.

sindresorhus: Yes, forbid.

PR #123, opened 2019-09-02
Add setting to indicate a flag can be set multiple times

### LINKED ISSUE #111: Add setting to indicate a flag can be set multiple times

ankeetmaini:
Fixes #111 

Taken inspiration and some test cases from #112

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -101,6 +101,7 @@ The key is the flag name and the value is an object with any of:
 - `type`: Type of value. (Possible values: `string` `boolean`)
 - `alias`: Usually used to define a short flag alias.
 - `default`: Default value when the flag is not specified.
+- `multiple`: To indicate that a flag can be passed multiple times, value will be an array. Default value is `false`
 
 Example:
 
@@ -110,6 +111,7 @@ flags: {
 		type: 'string',
 		alias: 'u',
 		default: 'rainbow'
+		multiple: true
 	}
 }
 ```
--- a/test.js
+++ b/test.js
@@ -22,7 +22,7 @@ test('return object', t => {
 	t.true(cli.flags.fooBar);
 	t.is(cli.flags.meow, 'dog');
 	t.is(cli.flags.unicorn, 'cat');
-	t.deepEqual(cli.flags['--'], ['unicorn', 'cake']);
+	t.deepEqual(cli.flags['--'], 'cake');
 	t.is(cli.pkg.name, 'meow');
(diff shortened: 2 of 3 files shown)


ankeetmaini:
ping @sindresorhus :)

sindresorhus:
You need to update index.d.ts too.

readme.md
```diff
@@ -101,6 +101,7 @@ The key is the flag name and the value is an object with any of:
 - `type`: Type of value. (Possible values: `string` `boolean`)
 - `alias`: Usually used to define a short flag alias.
 - `default`: Default value when the flag is not specified.
+- `multiple`: To indicate that a flag can be passed multiple times, value will be an array. Default value is `false`
```
sindresorhus:
```suggestion
- `multiple`: Indicates that a flag can be passed multiple times. The flag value will be an array. (Default: `false`)
```

test.js
```diff
@@ -240,3 +240,67 @@ test('disable autoVersion/autoHelp if `cli.input.length > 0`', t => {
...
+	t.deepEqual(meow({
+		flags,
+		argv: [
+			'--foo=a',
+			'--foo=b',
+			'--foo=c',
+			'--foo=d'
+		]
+	}).flags, {foo: ['a', 'b', 'c', 'd']});
+});
```
sindresorhus: Can you add tests to ensure it works correctly with `flags.type == 'boolean'` and `flags.type == 'number'`?

test.js
```diff
@@ -240,3 +240,67 @@ test('disable autoVersion/autoHelp if `cli.input.length > 0`', t => {
...
+
+test('combination of multi flag and single', t => {
+	const cli = meow({
+		argv: ['--foo=1', '--foo=2', '--bar', '1', '--foo', '3', '--bar=5', '-f', '4'],
+		flags: {
+			foo: {multiple: true, alias: 'f', type: 'string'},
+			bar: {type: 'string'}
+		}
+	});
+	t.deepEqual(cli.flags.foo, [1, 2, 3, 4]);
```
sindresorhus: Shouldn't the elements in the array be strings?

sindresorhus:
Bump :)

PR #132, opened 2019-12-11
Handle alias arrays generating unnormalized flags

### LINKED ISSUE #131: Handling arrays of aliases
When the `minimist-options` package parses aliases, it was apparently designed to handle both a single string as well as an array of strings ([it `arrify`s the alias and turns a single string into an array before processing](https://github.com/vadimdemedes/minimist-options/blob/f6d24087f871d6605b6c8d0f9c2b36af18467875/index.js#L62)). (This package was also written by @sindresorhus, but the [original `minimist` handled aliases similarly](https://github.com/substack/minimist/blob/4cf45a26b9af5f4ddab63107f4007485e576cfd3/index.js#L18).)

That is quite convenient if you want multiple aliases for a single option. 

However, it creates some inconsistent behavior in the latest version with `unnormalizedFlags`. If you have more than one alias for a flag, those items (the flag and all aliases) get included in `cli.flags` as well as `unnormalizedFlags`. If there's only one alias (either a single element array or a plain string), the normal behavior of the alias being excluded from `flags` is observed.

The ability to pass in an array into aliases is admittedly an undocumented, if not unsupported, feature, but for me it's quite helpful. In particular it helps with backwards compatibility and gradual deprecation of flags that have been renamed. I can support both the old and new flag names for 1-2 versions while warning of the deprecation, without having to rewrite the underlying code.

The current behavior is not difficult to deal with, once known. I am seeking to raise awareness of the handling of alias arrays and propose that it become a documented and supported feature.

benfletcher:
Minimist-options allows aliases to provided as an array. Until the introduction of `unnormalizedFlags` this did not cause any problems. The removal of aliases from the flag list silently fails if more than one alias is included for a particular flag.

Fixes #131

### DIFF
--- a/index.js
+++ b/index.js
@@ -114,8 +114,15 @@ const meow = (helpText, options) => {
 
 	if (options.flags !== undefined) {
 		for (const flagValue of Object.values(options.flags)) {
-			delete flags[flagValue.alias];
+			if (Array.isArray(flagValue.alias)) {
+				flagValue.alias.forEach(alias =>
+					delete flags[flagValue.alias]
+				);
+			} else {
+				delete flags[flagValue.alias];
+			}
 		}
+
 	}
 
 	return {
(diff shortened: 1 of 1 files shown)


sindresorhus:
Sounds good. Can you add a test and update the docs (readme and index.d.ts)?

sindresorhus:
Bump :)

PR #135, opened 2020-02-14
Replace `minimist-options` types with own ones

### LINKED ISSUE #134: Missing `array` in `flags.type` document
Hi,

Currently, the `flags.type` possible values are documented as `string`, `boolean`, and  `number`:

https://github.com/sindresorhus/meow/blob/5975fe618eb8dd70c06911d34ab52dd689a60616/index.d.ts#L11

Also on README:

https://github.com/sindresorhus/meow#flags

But the actual values by `minimist-options` include also `array`:

```ts
export type OptionType = 'string' | 'boolean' | 'number' | 'array';
```

https://github.com/vadimdemedes/minimist-options/blob/f6d24087f871d6605b6c8d0f9c2b36af18467875/index.d.ts#L3

Do these descriptions about `flags.type` need to include `array`?

ybiquitous:
Fix #134

See also:

https://github.com/vadimdemedes/minimist-options/blob/f6d24087f871d6605b6c8d0f9c2b36af18467875/index.d.ts

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -13,9 +13,9 @@ declare namespace meow {
 	type BooleanFlag = Flag<'boolean', boolean>;
 	type NumberFlag = Flag<'number', number>;
 
-	type BaseFlags = { [key: string]: StringFlag | BooleanFlag | NumberFlag };
+	type AnyFlags = { [key: string]: StringFlag | BooleanFlag | NumberFlag };
 
-	interface Options<Flags extends BaseFlags> {
+	interface Options<Flags extends AnyFlags> {
 		/**
 		Define argument flags.
 
@@ -172,7 +172,7 @@ declare namespace meow {
 		readonly hardRejection?: boolean;
 	}
 
-	type TypedFlags<Flags extends BaseFlags> = {
+	type TypedFlags<Flags extends AnyFlags> = {
 		[F in keyof Flags]: Flags[F] extends {type: 'number'}
 			? number
 			: Flags[F] extends {type: 'string'}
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1,8 +1,21 @@
 import {PackageJson} from 'type-fest';
-import {Options as MinimistOptions} from 'minimist-options';
 
 declare namespace meow {
-	interface Options<Flags extends MinimistOptions> {
+	type FlagType = 'string' | 'boolean' | 'number';
+
+	interface Flag<Type extends FlagType, Default> {
+		readonly type?: Type;
+		readonly alias?: string;
```
ybiquitous: Considering #132, we may change like this:

```suggestion
		readonly alias?: string | ReadonlyArray<string>;
```

index.d.ts
```diff
@@ -1,8 +1,21 @@
...
+		readonly default?: Default;
+	}
+
+	type StringFlag = Flag<'string', string>;
+	type BooleanFlag = Flag<'boolean', boolean>;
+	type NumberFlag = Flag<'number', number>;
+
+	type BaseFlags = { [key: string]: StringFlag | BooleanFlag | NumberFlag };
```
ybiquitous: I don't really like the name `BaseFlags`. Do you have another good name?

sindresorhus: `AnyFlags`?

index.d.ts
sindresorhus: That should be done in that PR though.

index.d.ts
ybiquitous: Looks nice! Thank you, I'll change.

index.d.ts
ybiquitous: Agreed. 👍

sindresorhus:
Thanks :)

ybiquitous:
You are welcome!

PR #82, opened 2018-03-16
Fix #80: add autoHelpFlags

matthewadams:
Automatically generate help for flags.  Fixes https://github.com/sindresorhus/meow/issues/80.

PR #140, opened 2020-03-22
Generate help by flag descriptions

### LINKED ISSUE #80: Autogenerate flag help when no helpMessage is given
I'm coming from `commander`, which autogenerates documentation on flags, and was surprised to see that `meow`'s `autoHelp` feature doesn't do the same.

```javascript
// in index.js
const cli = require('meow')({
  flags: {
    'initial-indent': {
      type: 'string',
      alias: 'n',
      description: 'Initial indentation level'
      default: '0'
    },
    indent: {
      type: 'string',
      alias: 'i',
      description: 'YAML indentation level'
      default: '2'
    }
  }
})
cli.showHelp()
```
should generate output something like
```shell
$ ./index.js

  My cool tool description

  Options:
    -n, --initial-indent  [value]  Initial indentation spaces.  Optional; default 0.
    -i, --indent          [value]  YAML indentation spaces.  Optional; default 2.
```

ybiquitous:
This fixes #80. Rechallenge of PR #82.

## Example

```
  A useful command

  Options:
    --foo <string>

    --output <string>

    --input <string>  (default: stdin)

    -i <number>, --indent <number>  (default: 2)
      Indent level

    --enabled  (default: false)
      Enabled or not

    --lw <string>, --long-word <string>  (default: none)
      Very very long long word.
      This is the second line.
```

Welcome any feedback!


test.js
```diff
@@ -35,6 +35,106 @@ test('support help shortcut', t => {
...
+	});
+	t.is(cli.help, `
+  CLI app helper
+
+  Options:
+    --foo <string>
+
+    --output <string>
+
+    --input <string>  (default: stdin)
```
ybiquitous: [question] Which style do you prefer?

```suggestion
    --input <string>  [default: 'stdin']
```

sindresorhus: `[default: stdin]`

index.js
```diff
@@ -86,6 +88,53 @@ const meow = (helpText, options) => {
 
 	help = (description ? `\n  ${description}\n` : '') + (help ? `\n${help}\n` : '\n');
 
+	if (options.helpOptions) {
```
sindresorhus, line 91: There's a lot of logic here. Most of the logic here should be moved into one or more functions.

test.js
```diff
@@ -35,6 +35,106 @@ test('support help shortcut', t => {
...
+
+    -i <number>, --indent <number>  (default: 2)
```
sindresorhus, line 105: The short-flag should be after the long-flag.

sindresorhus:
I don't think `helpOptions` is a good option name. I think we should just automatically show the text if the `help` option (or argument) is not specified.

sindresorhus:
Also look at how https://github.com/yargs/yargs solves this and how it presents the output.

sindresorhus:
There should probably also be a method to get the autogenerated help text in case you want to manually combine it with your own help text.

ybiquitous:
Sorry for my late reply. I will work on your review this weekend.

test.js
ybiquitous: For example, `node` or `git` puts the short-flag before the long-flag. Could you please the reason you recommend the long-flag?

```
$ node -h
Usage: node [options] [ script.js ] [arguments]
       node inspect [options] [ script.js | host:port ] [arguments]

Options:
  -                                         script read from stdin (default if no file name is provided,
                                            interactive mode if a tty)
  --                                        indicate the end of node options
  --abort-on-uncaught-exception             aborting instead of exiting causes a core file to be generated
                                            for analysis
  -c, --check                               syntax check script without executing
  --completion-bash                         print source-able bash completion script
  --cpu-prof                                Start the V8 CPU profiler on start up, and write the CPU profile
...
```

https://git-scm.com/docs/git-commit#_options

index.js
ybiquitous: Make sense. 👍

ybiquitous:
I looked at some command-line helpers and some command-line tools. For example:

- [yargs/yargs](https://github.com/yargs/yargs)
- [Ruby's OptionParser](https://ruby-doc.org/stdlib-2.7.1/libdoc/optparse/rdoc/OptionParser.html)
- [Python's argparse](https://docs.python.org/3.8/library/argparse.html)
- `node` command of Node.js

As a result, I think the following format is better:

```
An awesome CLI tool

Usage: foo [options]

Options:
  --input <string>            Input source. [default: stdin]
  -o, --output <string>       Output file to be reported.
  -i, --indent <number>       Indent level. [default: 2]
  --quiet                     Suppress any output.
  --verbose                   Show verbose output.
                              See also --quiet.
  -h, --help                  Show help.
```

- An option's name and description on a line.
- First short-flag, last long-flag (all tools except `yargs` support this order).

sindresorhus:
I agree with everything except the shortflag order. The shortflag should be last.

sindresorhus:
> Indent level. [default: 2]

Do any of the libraries put more than one space before `[default`? It feels a bit cramped into the description.

sindresorhus:
Another good library for inspiration: https://github.com/apple/swift-argument-parser

ybiquitous:
> I agree with everything except the shortflag order. The shortflag should be last.

OK, I'll do it. 👍 

> Do any of the libraries put more than one space before `[default`? It feels a bit cramped into the description.

I have no strong opinion about it. How many spaces are appropriate?

sindresorhus:
2

test/test.js
```diff
@@ -31,14 +31,6 @@ test('return object', t => {
...
 });
 
-test('support help shortcut', t => {
-	const cli = meow(`
-		unicorn
-		cat
-	`);
-	t.is(cli.help, indentString('\nCLI app helper\n\nunicorn\ncat\n', 2));
-});
-
```
ybiquitous, line 41: Moved to `test/help.js`.

ybiquitous:
Sorry for the late response... I've addressed all the review points except for the following one:

> There should probably also be a method to get the autogenerated help text in case you want to manually combine it with your own help text.

About the comment above, what kind of API do you expect?

generate-help.js
```diff
@@ -0,0 +1,93 @@
...
+		const fullPadding = ' '.repeat(maxNameLengh + spaces);
+		restLines = restLines.map(line => fullPadding + line).join(EOL);
+
+		return `${name}${padding}${firstLine}${EOL}${restLines}`;
+	});
+
+	return lines;
+}
+
+module.exports = function ({description, help, flags}, pkg) {
```
sindresorhus, line 63:
```suggestion
module.exports = ({description, help, flags}, pkg) => {
```

generate-help.js
```diff
@@ -0,0 +1,93 @@
...
+			lines.push('');
+		}
+
+		lines.push('Options:');
+		lines.push(...flagsSection(flags).map(line => redent(line, 2)));
+	}
+
+	lines = lines.map(line => trimNewlines(line));
+
+	const content = lines.join(EOL).replace(/^\t+/gm, '').replace(/[\t ]+[\r\n]*$/gm, '');
```
sindresorhus, line 91: Instead of the regexes here, couldn't you use `redent`?

generate-help.js
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		entry.desc = entry.desc.trim();
+
+		return entry;
+	});
+
+	const maxNameLengh = Math.max(...entries.map(({name}) => name.length));
+
+	const lines = entries.map(({name, desc}) => {
```
sindresorhus, line 41: Don't use abbreviations for variable/parameter names.

generate-help.js
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		lines.push('Options:');
+		lines.push(...flagsSection(flags).map(line => redent(line, 2)));
+	}
+
+	lines = lines.map(line => trimNewlines(line));
+
+	const content = lines.join(EOL).replace(/^\t+/gm, '').replace(/[\t ]+[\r\n]*$/gm, '');
+	return EOL + trimNewlines(redent(content, 2)) + EOL;
```
sindresorhus, line 92: Use `\n` instead of EOL

sindresorhus:
> About the comment above, what kind of API do you expect?

Any suggestions?

ybiquitous:
> > About the comment above, what kind of API do you expect?
> 
> Any suggestions?

For example:

```js
const { generateHelp } = require("meow");
generateHelp({ help: "...", description: "...", flags: {...} }, packageJson);
```

ybiquitous:
Edited:

```js
const { generateHelp } = require("meow");
generateHelp({ help: "...", description: "...", defaultDescription: "...", flags: {...} });
```

sindresorhus:
Thinking more about it, maybe instead of a free method, we could make the `help` option also accept a function, which would receive the full generated help text, generated each section (for example, the rendered help for flags, and other sections), and also the meow options. That should be flexible enough for the user to do anything they want. The function would be expected to return the final `help` text. Thoughts?

ybiquitous:
Thanks for your advice.

> make the help option also accept a function

It seems a better API than my suggestion! 👍 
Do you expect the following use case, for example?

```ts
const cli = meow({
  help: (wholeText: string, flagsSection: string, description: string, options: meow.Options) => {
    return `...`;
  },
  flags: {
    input: {
      type: 'string',
      description: 'Input file path'
    }
  }
});
```

sindresorhus:
Yes, but I think it’s better to make `help` accept a single object as argument instead of lots of arguments. Since many of them will probably not be used.

ybiquitous:
Make sense. I'll try it, thanks!

ybiquitous:
The `help` option can receive a function via 935911f.

ybiquitous:
I'm ready to be reviewed again.

index.d.ts
```diff
@@ -27,6 +28,11 @@ declare namespace meow {
 
 	type AnyFlags = {[key: string]: StringFlag | BooleanFlag | NumberFlag};
 
+	/**
+	Callback function to generate a help text you want.
+	*/
+	type GenerateHelp = (args: Readonly<{wholeText: string; flagLines: readonly string[]; description: string; options: Readonly<Options<AnyFlags>>}>) => string;
```
sindresorhus, line 34: The `args` type should be a separate `interface` and each property should have a doc-comment.

sindresorhus: `args` should have a better name too.

index.d.ts
```diff
@@ -27,6 +28,11 @@ declare namespace meow {
 
 	type AnyFlags = {[key: string]: StringFlag | BooleanFlag | NumberFlag};
 
+	/**
+	Callback function to generate a help text you want.
```
sindresorhus, line 32: This needs to be more verbose. Document the default behavior, etc.

readme.md
```diff
@@ -175,14 +177,27 @@ Set it to `false` to disable it altogether.
...
 
 The input is reindented and starting/ending newlines are trimmed which means you can use a [template literal](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/template_strings) without having to care about using the correct amount of indent.
 
 The description will be shown above your help text automatically.
 
+Also, you can customize the auto-generated help text by giving the function as follows:
+
+```js
+meow({
+  help: ({wholeText, flagLines, description, options}) => {
```
sindresorhus, line 192: Use tab-indentation.

sindresorhus:
Some things to keep in mind:

- Review your own diff.
- Try to refactor and simplify the implementation.
- Try to add more tests.
- More docs.

sindresorhus:
Bump

ybiquitous:
Sorry for the late response.

> * Try to refactor and simplify the implementation.
> * Try to add more tests.

I have no good ideas about the points above. Could you please help me?

generate-help.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
```
sindresorhus, line 1: This file needs to be added to `files` in package.json.

generate-help.js
```diff
@@ -0,0 +1,103 @@
...
+		result += ` <${type}>`;
+	}
+
+	return result;
+}
+
+function buildFlagLines(flags) {
+	flags = {...flags, help: {type: 'boolean', description: 'Show help'}};
+
+	const entries = Object.entries(decamelizeKeys(flags, '-')).map(([name, def]) => {
```
sindresorhus, line 22: Don't use abbreviations (regarding `def`, and other cases).

index.d.ts
```diff
@@ -28,6 +29,37 @@ declare namespace meow {
...
+		flagLines: readonly string[];
+
+		/**
+		An object including each flag information.
+		*/
+		flagOptions: Readonly<Options<AnyFlags>>;
+	}
+
+	/**
+	Callback function to customize a help text you want.
```
sindresorhus, line 55: Typos

sindresorhus (changes requested):

readme.md
```diff
@@ -175,14 +177,27 @@ Set it to `false` to disable it altogether.
...
+Type: `string | boolean | Function`
 
-The help text you want shown.
+The help text you want to show.
 
 The input is reindented and starting/ending newlines are trimmed which means you can use a [template literal](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/template_strings) without having to care about using the correct amount of indent.
 
 The description will be shown above your help text automatically.
 
+Also, you can customize the auto-generated help text by giving the function as follows:
```
sindresorhus, line 188: The readme and index.d.ts should be in sync.

sindresorhus:
> Try to refactor and simplify the implementation.

The code is quite hard to follow. Some code comments could maybe help that too. I often comment "simplify" on PRs as honestly most PRs could be simplified. Nobody writes perfect code the first time around. It helps to come back to it and look it over and try to simplify/refactor it. That's what I'm asking.

> Try to add more tests.

I'm just asking you to look over and try to add some more tests. I'm the one that will be stuck maintaining this for infinity, so I want to ensure it's as solid as possible.

index.d.ts
```diff
@@ -28,6 +29,37 @@ declare namespace meow {
 	type AnyFlag = StringFlag | BooleanFlag | NumberFlag;
 	type AnyFlags = {[key: string]: AnyFlag};
 
+	interface GenerateHelpOptions {
+		/**
+		A whole help text generated by default.
```
sindresorhus, line 34: Typo

index.d.ts
```diff
@@ -28,6 +29,37 @@ declare namespace meow {
...
+		*/
+		wholeText: string;
+
+		/**
+		A description of a command generated by default.
+		*/
+		description: string;
+
+		/**
+		A list of a line including each flag's name and description.
```
sindresorhus, line 44: Typo

index.d.ts
```diff
@@ -28,6 +29,37 @@ declare namespace meow {
...
+		*/
+		flagLines: readonly string[];
+
+		/**
+		An object including each flag information.
```
sindresorhus, line 49: Not clear to the user what "flag information" means.

sindresorhus:
This change needs more docs. There's not even a mention of the new behavior of generating flag docs by default.

armano2:
as for yargs, this is not fully true as they print in with a little better indentations

example:
```
@commitlint/cli@11.0.0 - Lint your commit messages

[input] reads from stdin if --edit, --env, --from and --to are omitted

Options:
  -c, --color          toggle colored output           [boolean] [default: true]
  -g, --config         path to the config file                          [string]
      --print-config   print resolved config          [boolean] [default: false]
  -d, --cwd            directory to execute in
                                         [string] [default: (Working Directory)]
      --edit           read last commit message from the specified file or
                       fallbacks to ./.git/COMMIT_EDITMSG               [string]
  -E, --env            check message in the file at path given by environment
                       variable value                                   [string]
  -x, --extends        array of shareable configurations to extend       [array]
  -H, --help-url       help url in error message                        [string]
  -f, --from           lower end of the commit range to lint; applies if
                       edit=false                                       [string]
  -o, --format         output format of the results                     [string]
  -p, --parser-preset  configuration preset to use for
                       conventional-commits-parser                      [string]
  -q, --quiet          toggle console output          [boolean] [default: false]
  -t, --to             upper end of the commit range to lint; applies if
                       edit=false                                       [string]
  -V, --verbose        enable verbose output for reports without problems
                                                                       [boolean]
  -v, --version        display version information                     [boolean]
  -h, --help           Show help                                       [boolean]
```

ybiquitous:
I am very sorry for my too late response. I can no longer go any further with this PR due to my lack, in spite of many advice I have received. 🙇 

I am so grateful for the kind review.
And, I hope this PR will be of some help when someone works on implementing this feature again. ☺️

Thank you.

PR #142, opened 2020-03-25
Update readme.md

konuralpt:
fixed styling issue


sindresorhus:
This is not correct. What's there is the intended look. This changes the meaning.

PR #143, opened 2020-04-11
Add `isMultiple` option for flags

### LINKED ISSUE #111: Add setting to indicate a flag can be set multiple times

ulken:
Differs in two ways from previous attempts:

1. Instead of setting the last value if a flag is used multiple times without `multiple` set, it throws an error. Makes more sense to me and is how I interpret _"we should **allow** only one value by default"_.
2. It supports `boolean` and `number` arrays in addition to `string`.

Fixes #111
Closes #123

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 yarn.lock
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 node_modules
 yarn.lock
-.vscode
--- a/index.d.ts
+++ b/index.d.ts
@@ -177,12 +177,12 @@ declare namespace meow {
 
 	type TypedFlags<Flags extends AnyFlags> = {
 		[F in keyof Flags]: Flags[F] extends {type: 'number'}
-		? number
-		: Flags[F] extends {type: 'string'}
-		? string
-		: Flags[F] extends {type: 'boolean'}
-		? boolean
-		: unknown;
+			? number
+			: Flags[F] extends {type: 'string'}
+				? string
+				: Flags[F] extends {type: 'boolean'}
+					? boolean
+					: unknown;
 	};
 
 	interface Result<Flags extends AnyFlags> {
--- a/index.d.ts
+++ b/index.d.ts
@@ -25,7 +25,7 @@ declare namespace meow {
 		- `type`: Type of value. (Possible values: `string` `boolean` `number`)
 		- `alias`: Usually used to define a short flag alias.
 		- `default`: Default value when the flag is not specified.
-		- `multiple`: Indicates a flag can be set multiple times. Returns an array. (Default: false)
+		- `multiple`: Indicates a flag can be set multiple times. Values are turned into an array. (Default: false)
 
 		@example
 		```
@@ -177,12 +177,12 @@ declare namespace meow {
 
 	type TypedFlags<Flags extends AnyFlags> = {
 		[F in keyof Flags]: Flags[F] extends {type: 'number'}
-			? number
-			: Flags[F] extends {type: 'string'}
-				? string
-				: Flags[F] extends {type: 'boolean'}
-					? boolean
-					: unknown;
+		? number
+		: Flags[F] extends {type: 'string'}
+		? string
+		: Flags[F] extends {type: 'boolean'}
+		? boolean
+		: unknown;
 	};
 
 	interface Result<Flags extends AnyFlags> {
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,6 +7,7 @@ declare namespace meow {
 		readonly type?: Type;
 		readonly alias?: string;
 		readonly default?: Default;
+		readonly multiple?: boolean;
 	}
 
 	type StringFlag = Flag<'string', string>;
@@ -24,6 +25,7 @@ declare namespace meow {
 		- `type`: Type of value. (Possible values: `string` `boolean` `number`)
 		- `alias`: Usually used to define a short flag alias.
 		- `default`: Default value when the flag is not specified.
+		- `multiple`: Indicates a flag can be set multiple times. Returns an array. (Default: false)
 
 		@example
 		```
@@ -31,7 +33,8 @@ declare namespace meow {
 			unicorn: {
 				type: 'string',
 				alias: 'u',
-				default: 'rainbow'
+				default: ['rainbow', 'cat'],
+				multiple: true
 			}
 		}
 		```
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,7 +7,7 @@ declare namespace meow {
 		readonly type?: Type;
 		readonly alias?: string;
 		readonly default?: Default;
-		readonly multiple?: boolean;
+		readonly isMultiple?: boolean;
 	}
 
 	type StringFlag = Flag<'string', string>;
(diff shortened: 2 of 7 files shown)


index.js
```diff
@@ -40,39 +42,57 @@ const meow = (helpText, options) => {
...
 		};
 	}
 
+	if (minimistOptions.array !== undefined) {
+		minimistOptions.array = arrify(minimistOptions.array).map(flagKey => ({
+			key: flagKey,
+			[options.flags[flagKey].type || 'string']: true
+		}));
+	}
+
```
ulken, line 93: This should preferably be supported by `minimist-options`, but I can't really see how as it uses `type: 'array'` and thus lose the sub-typing (boolean, number etc.).

ulken:
Extracted unrelated change into separate PR: #145

index.d.ts
```diff
@@ -24,14 +25,16 @@ declare namespace meow {
 		- `type`: Type of value. (Possible values: `string` `boolean` `number`)
 		- `alias`: Usually used to define a short flag alias.
 		- `default`: Default value when the flag is not specified.
+		- `multiple`: Indicates a flag can be set multiple times. Returns an array. (Default: false)
```
sindresorhus, line 28: *What* returns an array? It's not clear enough.

index.js
```diff
@@ -71,6 +84,13 @@ const meow = (helpText, options) => {
 		};
 	}
 
+	if (minimistOptions.array !== undefined) {
+		minimistOptions.array = arrify(minimistOptions.array).map(flagKey => ({
```
sindresorhus, line 88: Can you add a short code comments what's happening here?

index.js
```diff
@@ -112,10 +132,12 @@ const meow = (helpText, options) => {
 	const flags = camelcaseKeys(argv, {exclude: ['--', /^\w$/]});
 	const unnormalizedFlags = {...flags};
 
-	if (options.flags !== undefined) {
-		for (const flagValue of Object.values(options.flags)) {
-			delete flags[flagValue.alias];
+	for (const [flagKey, flagValue] of Object.entries(options.flags).filter(([key]) => key !== '--')) {
+		if (!flagValue.multiple && Array.isArray(flags[flagKey])) {
+			throw new Error(`Only one value allowed for --${flagKey}.`);
```
sindresorhus, line 137:
```suggestion
			throw new Error(`The flag \`--${flagKey}\` can only be set once.`);
```

?

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 yarn.lock
+.vscode
```
sindresorhus, line 3: This should be in your own global gitignore.

sindresorhus:
The main function is getting huge. Would be great if you could try to modularize some of the new code into separate functions.

sindresorhus:
Should it be named `isMultiple` instead of `multiple`? Since there will soon be a `isRequired` option (https://github.com/sindresorhus/meow/issues/131).

.gitignore
ulken: Fair enough

index.d.ts
ulken: Valid point. Maybe drop the word `return` altogether?

How about `Values are turned into an array (in order?)`?

index.js
ulken: Sure

index.js
ulken: 👍

ulken:
> The main function is getting huge. Would be great if you could try to modularize some of the new code into separate functions.

Sure, I'll give it a go.

ulken:
> Should it be named `isMultiple` instead of `multiple`? Since there will soon be a `isRequired` option (#131).

Personally, I tend to use `is/has/should` etc. when it's a function, but not for simple boolean values.

Having that said, I definitely think naming of booleans should follow the same convention. So if `isRequired` is set in stone, I think this should follow suit.

index.js
ulken: Better yet: Poke @vadimdemedes to give his input on supporting this in `minimist-options`: [`minimist-options` | Support boolean and number arrays](https://github.com/vadimdemedes/minimist-options/issues/18). Then there wouldn't be any need for a comment (or code!) at all...

Your input would be highly valuable as well!

index.d.ts
ulken: Went with the shorter form

ulken:
@sindresorhus Alright, I've processed your feedback.

ulken:
Up-to-date with `master` and reverted unintentional formatting change

test.js
```diff
@@ -307,3 +307,146 @@ test('supports `number` flag type - throws on incorrect default value', t => {
...
+				type: 'number',
+				isMultiple: true
+			}
+		}
+	}).flags, {
+		foo: [1.3, -1]
+	});
+});
+
+test('isMultiple - flag default values', t => {
```
sindresorhus, line 427: Can you add a test like this, but where the flag has multiple default values? `default: ['foo', 'bar']`?

ulken: Sure

ulken: Done

ulken:
@sindresorhus Up-to-date with `master` again...

sindresorhus:
Looks great 👌

ulken:
Cool. Would you be able to ask @vadimdemedes to check out the PR in `minimist-options` if you talk to him? Won’t respond on GitHub nor Twitter. 

Understand he might not want to merge it, just want it to be evaluated.

sindresorhus:
Give it a little bit more time. He's probably busy. Sometimes it takes some weeks for me to look at PRs too.

ulken:
Hehe, yeah, so I’ve learned 😉

I get that, but he hasn’t responded to a related issue from August (I think) last year. 

Oh, well. Thanks anyway.

sindresorhus:
I tried this on a project and it crashes in:

```
/Users/sindresorhus/dev/oss/capture-website-cli/node_modules/meow/index.js:86␊
    		[flags[flagKey].type || 'string']: true␊
    		                ^␊
    ␊
    TypeError: Cannot read property 'type' of undefined␊
        at arrify.map.flagKey (/Users/sindresorhus/dev/oss/capture-website-cli/node_modules/meow/index.js:86:19)␊
        at Array.map (<anonymous>)␊
        at convertToTypedArrayOption (/Users/sindresorhus/dev/oss/capture-website-cli/node_modules/meow/index.js:84:22)␊
        at meow (/Users/sindresorhus/dev/oss/capture-website-cli/node_modules/meow/index.js:148:25)␊
        at Object.<anonymous> (/Users/sindresorhus/dev/oss/capture-website-cli/cli.js:9:13)␊
        at Module._compile (internal/modules/cjs/loader.js:778:30)␊
        at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)␊
        at Module.load (internal/modules/cjs/loader.js:653:32)␊
        at tryModuleLoad (internal/modules/cjs/loader.js:593:12)␊
        at Function.Module._load (internal/modules/cjs/loader.js:585:3)␊
```

When the flag is `--hide-elements`. I think it doesn't correctly account for camel-casing.

ulken:
Ouch. OK, I’ll have a looksie later tonight.

ulken:
@sindresorhus Pushed fix

vadimdemedes:
Excuse me for delay, guys. Change is in master and published!

ulken:
https://github.com/sindresorhus/meow/commit/629af48c24a2f19636512b2f532cfcd9419f9be1 😉

PR #147, opened 2020-04-27
Document ES Modules usage

### LINKED ISSUE #130: Support ES module scripts?
I've tried writing a silly deploy script as an ES module, now that node has experimental support, but meow doesn't grok this:

```
> yarn deploy
yarn run v1.21.0
$ node src/__scripts__/deploy.mjs
(node:5588) ExperimentalWarning: The ESM module loader is experimental.
myproj/node_modules/meow/index.js:15
const parentDir = path.dirname(module.parent.filename);
                                             ^

TypeError: Cannot read property 'filename' of undefined
    at Object.<anonymous> (myproj/node_modules/meow/index.js:15:46)
```

I'm curious of meow really needs to look up the parent module?
If it's part of meow's magic, is there perhaps a work-around?

In my case, my script is a tiny part of a larger project, and at present I can't / don't want to invoke meow on the top-level thing. Instead I'm trying to write a tiny script that only happens to be the directory structure of `myproj`.

Is `meow` not a good fit here, or am I just missing something?

jimthedev:
Closes #130.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+estest/index.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - '14'
   - '12'
   - '10'
   - '8'
--- a/estest/index.js
+++ b/estest/index.js
@@ -1,0 +1,25 @@
+import { createRequire } from "module";
+
+const meow = createRequire(import.meta.url)("../index.js");
+
+meow(
+	`
+Usage
+  $ estest <input>
+
+Options
+  --rainbow, -r  Include a rainbow
+
+Examples
+  $ estest unicorns --rainbow
+  🌈 unicorns 🌈
+`,
+	{
+		flags: {
+			rainbow: {
+				type: "boolean",
+				alias: "r"
+			}
+		}
+	}
+);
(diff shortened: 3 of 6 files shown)


sindresorhus:
Can you also add Node.js 14 here: https://github.com/sindresorhus/meow/blob/c67d9f4f8813d03a97637d8460274e63956cd096/.travis.yml#L3

jimthedev:
@sindresorhus added 14 to Travis, also made the esm test conditional on node 14 and greater since otherwise it’ll throw an error. Let me know if there’s a different way you’d prefer to handle that one.

jimthedev:
Ok RTG, @sindresorhus! Thanks for this project and all your projects.

jimthedev:
Just tried this an ran into an issue. Don't merge for now.

.eslintignore
```diff
@@ -0,0 +1 @@
+estest/index.js
```
jimthedev, line 1: We must ignore this file due to https://github.com/eslint/eslint/issues/13133

jimthedev:
Ok I updated this one (which is now ready to review @sindresorhus). According to https://twitter.com/MylesBorins/status/1255950295806488576 it looks like basically the user must use createRequire if they want to keep using this in ES modules (but they are doing so using commonjs). Seems ok for now?

I am open to other ideas on how to accomplish this but at least it would unblock people looking to use it in their ES based cli's.

sindresorhus:
Thanks :)

PR #148, opened 2020-05-02
fix: update yargs-parser to 18.1.2 to fix "vulnerability" in 18.1.1

bnb:
"fixes" a low-impact security vulnerability introduces by yargs-parser as asserted by npm audit.

ref: https://www.npmjs.com/advisories/1500


sindresorhus:
I don't mind updating, but be aware that Meow is not affected by this as Meow doesn't support nested properties in CLI flags.

PR #150, opened 2020-05-10
Fix `isMultiple` not handling multi-word flags

ulken:
Fix for https://github.com/sindresorhus/meow/pull/143#issuecomment-626287226

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 const path = require('path');
 const buildParserOptions = require('minimist-options');
 const yargs = require('yargs-parser');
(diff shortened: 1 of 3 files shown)


ulken:
It should be noted that if/when https://github.com/vadimdemedes/minimist-options/pull/19 is merged, this won't be a problem anymore. But not knowing if/when that'll happen, I'd say it's best to fix this ASAP.

sindresorhus:
Thanks for fixing this quickly 🙌

PR #154, opened 2020-06-18
Improve flags types to acknowledge `isMultiple` and `isRequired` options

voxpelli:
This PR fixes two things in the types of flags:

1. Without this PR the type of an `isMultiple` flag is still reported to be a non-array.
2. All types are right now reported as always being defined, even when both `isRequired` and `default` is missing. This PR adds `|undefined` in those cases + does so in the case where `isRequired` is a function, as it's kind of hard to figure out what that function will return.

As a separate commit this PR also mirrors the Typescript config from `tsd` + extends it slightly. It does this for two reasons:

1. With a `tsconfig.json` in the project tools like VS Code can automatically configure itself correctly and provide correct feedback. Without the `tsconfig.json` it complained about some of the definition tests.
2. To be able to test `|undefined` one needs to have `strictNullChecks` set, else Typescript will just ignore any `undefined` added in the types.

I also added in `noImplicitAny` and `noUnusedLocals` in the `tsconfig.json` as both of those are helpful and good practice as well.

**Somewhat breaking:**

The addition of `|undefined` to all types that are missing both `isRequired` and `default` can cause some complaints in the code for those who have `strictNullChecks` set to `true`. For everyone else it will cause no change.

These complaints are correct though, but might cause some headaches for eg. those without lock-files, where it can have their CI fail.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -205,17 +205,17 @@ declare namespace meow {
 					? boolean
 					: unknown;
 
-	type PossiblyOptionalFlag<Flag extends AnyFlag> =
+	type PossiblyOptionalFlag<Flag extends AnyFlag, FlagType> =
 		Flag extends {isRequired: true}
-			? TypedFlag<Flag>
+			? FlagType
 			: Flag extends {default: any}
-				? TypedFlag<Flag>
-				: TypedFlag<Flag>|undefined;
+				? FlagType
+				: FlagType|undefined;
 
 	type TypedFlags<Flags extends AnyFlags> = {
 		[F in keyof Flags]: Flags[F] extends {isMultiple: true}
-			? Array<TypedFlag<Flags[F]>>
-			: PossiblyOptionalFlag<Flags[F]>
+			? PossiblyOptionalFlag<Flags[F], Array<TypedFlag<Flags[F]>>>
+			: PossiblyOptionalFlag<Flags[F], TypedFlag<Flags[F]>>
 	};
 
 	interface Result<Flags extends AnyFlags> {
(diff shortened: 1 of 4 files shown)


voxpelli:
Fixed the failing tests 🙈 (A downside of being used to `husky`, one forgets to manually run tests before push)

sindresorhus:
@NiGhTTraX Would you be able to give this a quick look?

index.d.ts
```diff
@@ -195,14 +196,26 @@ declare namespace meow {
...
+	type PossiblyOptionalFlag<Flag extends AnyFlag> =
+		Flag extends {isRequired: true}
+			? TypedFlag<Flag>
+			: Flag extends {default: any}
+				? TypedFlag<Flag>
+				: TypedFlag<Flag>|undefined;
+
+	type TypedFlags<Flags extends AnyFlags> = {
+		[F in keyof Flags]: Flags[F] extends {isMultiple: true}
+			? Array<TypedFlag<Flags[F]>>
```
NiGhTTraX, line 217: I think this branch should have the same logic as in `PossiblyOptionalFlag`. A flag with `isMultiple: true` but no `default` will return `undefined` for an empty argv, no? I couldn't find a test covering this case, but tried the following locally and it fails:

```
t.deepEqual(meow({
	argv: [],
	flags: {
		fooBar: {
			type: 'string',
			isMultiple: true
		}
	}
}).flags, {
	fooBar: []
});
```

```
  Difference:

    {
  +   fooBar: [],
    }
```

Maybe this could be addressed in another PR, alongside with a test, since this already seems like an improvement.

NiGhTTraX:
LGTM otherwise.

index.d.ts
sindresorhus: @voxpelli

voxpelli: Thanks for the ping @sindresorhus, pushed a fix

PR #159, opened 2020-08-28
Fix compatibility with `vercel/ncc`

Idered:
related #67

### DIFF
--- a/index.js
+++ b/index.js
@@ -153,7 +153,7 @@ const meow = (helpText, options) => {
 
 	const showVersion = () => {
 		console.log(typeof options.version === 'string' ? options.version : pkg.version);
-		process.exit();
+		process.exit(0);
 	};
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -12,7 +12,7 @@ const normalizePackageData = require('normalize-package-data');
 
 // Prevent caching of this module so module.parent is always accurate
 delete require.cache[__filename];
-const parentDir = path.dirname(module.parent.filename);
+const parentDir = path.dirname(module.parent?.filename ?? '.');
```
sindresorhus, line 15: This syntax is not supported in the targeted Node.js version.

Idered: Sorry, that was quick patch made in browser. Fixed and tested locally.

Idered:
Ok, looks like it's failing on latest node. I'll take a look later

Idered:
@sindresorhus fixed 👌

PR #162, opened 2020-10-01
Make `isMultiple` non-greedy

### LINKED ISSUE #160: isMultiple string flag is matching also the input arguments
I have a flag (called `multiple`) with a string `type` and the `isMultiple` flag. When I use that flag with an input argument, the input argument is now part of the flag values. This doesn't happen if you put the flag at the end.

### Example code
Test repo: https://github.com/dfernandez79/ismultiple-flag-example

```js
const meow = require('meow');
const cli = meow('This is a test. Try --multiple=string inputArg', {
  flags: {
    multiple: {
      type: 'string',
      isMultiple: true,
    }
  }
});

console.log('Input: ', cli.input);
console.log('Multiple: ', cli.flags.multiple)
```

### Outputs
```
➜ node . --multiple=string inputArg
Input:  []
Multiple:  [ 'string', 'inputArg' ]
```

```
➜ node . inputArg --multiple=string
Input:  [ 'inputArg' ]
Multiple:  [ 'string' ]
```

ulken:
I.e don't consume more than one argument.

The default behavior of `yargs-parser` is to [greedily consume array flag arguments](https://github.com/yargs/yargs-parser#greedy-arrays).

I also realized it doesn't support comma-separated values. Not sure if we want that or not, but I'll make a separate PR and let you be the judge.

Fixes #160

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -136,7 +136,7 @@ declare namespace meow {
 
 		const cli = meow(`
 			Usage
-			  $ foo
+				$ foo
 
 			Options
 				--rainbow, -r  Include a rainbow
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,6 +40,8 @@ declare namespace meow {
 		- `isRequired`: Determine if the flag is required.
 			If it's only known at runtime whether the flag is required or not you can pass a Function instead of a boolean, which based on the given flags and other non-flag arguments should decide if the flag is required.
 		- `isMultiple`: Indicates a flag can be set multiple times. Values are turned into an array. (Default: false)
+			Multiple values are provided by specifying the flag multiple times, e.g. `$ foo -u rainbow -u cat`.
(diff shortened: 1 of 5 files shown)


sindresorhus:
I think we should explicitly document this behavior.

sindresorhus:
Bump :)

ulken:
Yeah, sorry. I started looking into it, but then my son woke up 😅

I'll get to it in the coming days.

ulken:
@sindresorhus Alright. All yours.

voxpelli:
LGTM 👍

ulken:
Came to think of: do we want the tests to be as explicit? We have a test for space-separation. Should I add a test for a comma-separated value as well?

ulken:
Went ahead added it. Feel to skip/revert.

sindresorhus:
Thanks for this 👍🏻

PR #163, opened 2020-10-01
Default `isMultiple` to empty array

### LINKED ISSUE #161: `isMultiple` should result in an empty array when the flag is not specified
Currently, it returns `undefined`, which is incorrect.

I had to work around it here: https://github.com/sindresorhus/replace-in-files-cli/commit/e9ac8819c054eda48e2b273c1ee31ff0164f0b08#diff-2cce40143051e25f811b56c79d619bf5R63

// @ulken

ulken:
Note: Changes behavior for `booleanDefault: undefined` as well. With this change in place, I don't think it makes sense to filter out unset boolean array flags any longer.

Fixes #161

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -131,7 +131,7 @@ declare namespace meow {
 
 		_Note: If used in conjunction with `isMultiple`, the default flag value is set to `[]`._
 
-		__Caution: Explicitly specifying undefined for `booleanDefault` has different meaning from omitting key itself.__
+		__Caution: Explicitly specifying `undefined` for `booleanDefault` has different meaning from omitting key itself.__
 
 		@example
 		```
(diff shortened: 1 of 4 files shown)


sindresorhus:
> Changes behavior for booleanDefault: undefined as well.

I think the behavior should be made clear in the readme & index.d.ts.

ulken:
@sindresorhus Gave it a go.

While I was at it, I took the liberty to

1. Change the structure to adhere to the rest of the **README**, i.e. break it into two paragraphs. First a sentence to describe what the flag does, followed by a more detailed description of its ins and outs.
2. Synced **README** and **index.d.ts** in terms of content (one line was missing from the **README**).

If you disapprove, let me know and I'll revert.

sindresorhus:
Looks good 👍🏻

Issue #165, opened 2020-10-05
isRequired does not recognize kebab-case flags


alexmcmanus:
The `isRequired` option does not seem to work when the flag is defined as kebab-case. I think it's probably looking for the kebab-case key in a map of camelCase flags when it checks to see if it has been provided.

Given this test fixture:
```js
const cli = meow({
	description: "Custom description",
	help: `
		Usage
		  foo <input>
  `,
	flags: {
		test: {
			type: "string",
			alias: "t",
			isRequired: true,
		},
		"kebab-case": {
			type: "string",
			isRequired: true,
		},
		number: {
			type: "number",
			isRequired: true,
		},
		notRequired: {
			type: "string",
		},
	},
});

console.log(`${cli.flags.test},${cli.flags.kebabCase},${cli.flags.number}`);
```

This test case fails:
```js
test("spawn cli and test specifying all required flags", async (t) => {
	const { stdout } = await execa(fixtureRequiredPath, [
		"-t",
		"test",
		"--kebab-case",
		"test",
		"--number",
		"6",
	]);
	t.is(stdout, "test,test,6");
});
```

The failure is:
```
Missing required flag␊
    	--kebab-case
```

sindresorhus:
The keys in the flags option are expected to be camel-case.

alexmcmanus:
Thanks, I didn't realize the code matches camelCase flag definitions with kebab-case arguments. The only (minor) consequence is that my help text documents the kebab-case flags, but the isRequired error complains about the missing flag using camelCase.

sindresorhus:
> I didn't realize the code matches camelCase flag definitions with kebab-case arguments

Sounds like we need to improve the docs for that.

> but the isRequired error complains about the missing flag using camelCase.

That is a bug.

// @sbencoding

alexmcmanus:
Thanks. Given the flag is missing, the only way of influencing how it is reported seems to be either defining the flag in the preferred case, or having some Meow option to control the case used for reporting. The former seems simpler to me, and generally Meow seems to work OK if the the flag keys are kebab-case (although probably they then wouldn't match against camelCase flags if that's how they were provided). 

Any thoughts on how to approach this? I'll contribute if I can.

sindresorhus:
Solution:
- [ ] Decamelize the flag from options before showing it in the error message: `camelCase` => `--camel-case`.
[...]

PR #166, opened 2020-10-05
isRequired does not recognize kebab-case flags

alexmcmanus:
Fix for #165 - the isRequired option is broken if the flag is defined in camel-case.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 const path = require('path');
 const buildParserOptions = require('minimist-options');
 const parseArguments = require('yargs-parser');
+const camelCase = require('camelcase');
 const camelCaseKeys = require('camelcase-keys');
 const decamelizeKeys = require('decamelize-keys');
 const trimNewlines = require('trim-newlines');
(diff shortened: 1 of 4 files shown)


sbencoding:
Hi!
I don't know if I should post this here or on the issue, but a shorter solution, that doesn't require an additional dependency would be to pass `argv` instead of `flags` when `getMissingRequiredFlags` is called, because `argv` has the kebab-case flags as well as the camelCase variations, but only the camelCase variation is available in `flags`.

alexmcmanus:
Thanks for your feedback. The new dependency is an existing transitive dependency made explicit, but I agree your solution is simpler. I'm happy to make the change, but I'm not too sure this general approach is acceptable to the project maintainer, as it seems flags are supposed to be defined in camelCase. I'll post another comment and see if we can agree an approach. Thanks again.

sindresorhus:
Closing in favor of https://github.com/sindresorhus/meow/issues/165#issuecomment-712394008

PR #167, opened 2020-10-20
Gracefully handle package.json not being found

ryanrhee:
`read-pkg-up`'s `sync()` [can return undefined](https://github.com/sindresorhus/read-pkg-up/blob/4406073/index.d.ts#L82-L84).
If that happens, we try to index into `.packageJson`, which fails.

Making the index operation null-safe makes execution continue on with `pkg` being set to an empty object, which is what seems to be the original intent.

### DIFF
--- a/index.js
+++ b/index.js
@@ -91,11 +91,14 @@ const meow = (helpText, options) => {
 		helpText = '';
 	}
 
+	const foundPkg = readPkgUp.sync({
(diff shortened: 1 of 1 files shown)


sindresorhus:
You cannot use this syntax. It's not supported in all the supported Node.js versions.

PR #169, opened 2020-11-27
Add `allowUnknownFlags` options

### LINKED ISSUE #126: Proposal: fail on extraneous flags
Running `git -q` results in:
```
unknown option: -q
usage: git [--version] [--help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | -P | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]
```
I think that is a good feature because it prevents the user from fat fingering.  

Regarding property name, it could be called `allowExtraneous`, just `extraneous` or something else. 

I would go as far as defaulting it to `false`, but I guess that would require a major release, let alone an acceptance of this feature.

EDIT: I can provide PR with tests.

weareoutman:
fix #126 

To simplify figuring out which flags are unknown, I tried:

1. Pass [unknown-options-as-args](https://www.npmjs.com/package/yargs-parser#unknown-options-as-args) to yargs-parser first;
2. Filter unknown options collected in `argv._` which starts with `-`.

And I also make `allowUnknownFlags` be `true` by default to keep compatibility.

Let me know if I missed something.

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,8 +55,10 @@ const reportMissingRequiredFlags = missingRequiredFlags => {
 };
 
 const reportUnknownFlags = unknownFlags => {
-	console.error(`Unknown flag${unknownFlags.length > 1 ? 's' : ''}`);
-	console.error(unknownFlags.join('\n'));
+	console.error([
+		`Unknown flag${unknownFlags.length > 1 ? 's' : ''}`,
+		...unknownFlags
+	].join('\n'));
 };
 
 const buildParserFlags = ({flags, booleanDefault}) => {
--- a/index.js
+++ b/index.js
@@ -54,6 +54,11 @@ const reportMissingRequiredFlags = missingRequiredFlags => {
 	}
 };
 
+const reportUnknownFlags = unknownFlags => {
+	console.error(`Unknown flag${unknownFlags.length > 1 ? 's' : ''}`);
+	console.error(unknownFlags.join('\n'));
+};
+
 const buildParserFlags = ({flags, booleanDefault}) => {
 	const parserFlags = {};
 
@@ -110,6 +115,7 @@ const meow = (helpText, options) => {
 		autoVersion: true,
 		booleanDefault: false,
 		hardRejection: true,
+		allowUnknownFlags: true,
 		...options
 	};
 
@@ -137,6 +143,11 @@ const meow = (helpText, options) => {
 
 	if (parserOptions['--']) {
 		parserOptions.configuration['populate--'] = true;
+	}
+
+	if (!options.allowUnknownFlags) {
+		// Collect unknown options in `argv._` to be checked later.
+		parserOptions.configuration['unknown-options-as-args'] = true;
 	}
 
 	const {pkg} = options;
@@ -176,6 +187,14 @@ const meow = (helpText, options) => {
 
 	const input = argv._;
 	delete argv._;
+
+	if (!options.allowUnknownFlags) {
+		const unknownFlags = input.filter(item => typeof item === 'string' && item.startsWith('-'));
+		if (unknownFlags.length > 0) {
+			reportUnknownFlags(unknownFlags);
+			process.exit(2);
+		}
+	}
 
 	const flags = camelCaseKeys(argv, {exclude: ['--', /^\w$/]});
 	const unnormalizedFlags = {...flags};
--- a/test/allow-unkonwn-flags.js
+++ b/test/allow-unkonwn-flags.js
@@ -6,11 +6,13 @@ const fixtureAllowUnknownFlags = path.join(__dirname, 'fixtures', 'fixture-allow
 
 test('spawn CLI and test specifying unknown flags', async t => {
 	const error = await t.throwsAsync(
-		execa(fixtureAllowUnknownFlags, ['--foo', 'bar', '--unspecified-a', '--unspecified-b', 'input-is-allowed'])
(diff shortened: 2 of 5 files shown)


test/allow-unkonwn-flags.js
```diff
@@ -0,0 +1,23 @@
+import test from 'ava';
+import execa from 'execa';
+const path = require('path');
```
sindresorhus, line 3:
```suggestion
import path from 'path';
import test from 'ava';
import execa from 'execa';
```

test/allow-unkonwn-flags.js
```diff
@@ -0,0 +1,23 @@
...
+
+const fixtureAllowUnknownFlags = path.join(__dirname, 'fixtures', 'fixture-allow-unknown-flags.js');
+
+test('spawn cli and test specifying unknown flags', async t => {
```
sindresorhus, line 7:
```suggestion
test('spawn CLI and test specifying unknown flags', async t => {
```

test/allow-unkonwn-flags.js
```diff
@@ -0,0 +1,23 @@
...
+	try {
+		await execa(fixtureAllowUnknownFlags, ['--foo', 'bar', '--unspecified-a', '--unspecified-b', 'input-is-allowed']);
```
sindresorhus, line 9: Use `t.throwsAsync`. It returns an error that can be tested further.

test/allow-unkonwn-flags.js
```diff
@@ -0,0 +1,23 @@
...
+		const {stderr, message} = error;
+		t.regex(message, /Command failed with exit code 2/);
+		t.regex(stderr, /Unknown flag/);
+		t.regex(stderr, /--unspecified-a/);
+		t.regex(stderr, /--unspecified-b/);
+		t.notRegex(stderr, /input-is-allowed/);
+	}
+});
+
+test('spawn cli and test specifying known flags', async t => {
```
sindresorhus, line 20:
```suggestion
test('spawn CLI and test specifying known flags', async t => {
```

sindresorhus (changes requested):

index.js
```diff
@@ -54,6 +54,13 @@ const reportMissingRequiredFlags = missingRequiredFlags => {
 	}
 };
 
+const reportUnknownFlags = unknownFlags => {
+	console.error(`Unknown flag${unknownFlags.length > 1 ? 's' : ''}`);
+	for (const flag of unknownFlags) {
+		console.error(flag);
+	}
```
sindresorhus, line 61: It's better to only call `console.error` once for the whole output.

```js
unknownFlags.join('\n');
```

weareoutman: Thanks, all fixed ☕️

sindresorhus:
In the future, don't force push. It makes it impossible for me to see what changed, and I'm forced to re-review all the code.

test/allow-unkonwn-flags.js
```diff
@@ -0,0 +1,22 @@
+import path from 'path';
+import test from 'ava';
+import execa from 'execa';
+
+const fixtureAllowUnknownFlags = path.join(__dirname, 'fixtures', 'fixture-allow-unknown-flags.js');
+
+test('spawn CLI and test specifying unknown flags', async t => {
+	const error = await t.throwsAsync(
+		execa(fixtureAllowUnknownFlags, ['--foo', 'bar', '--unspecified-a', '--unspecified-b', 'input-is-allowed'])
+	);
```
sindresorhus, line 10: The `message` can be tested in `t.throwsAsync` directly.

index.js
```diff
@@ -54,6 +54,11 @@ const reportMissingRequiredFlags = missingRequiredFlags => {
 	}
 };
 
+const reportUnknownFlags = unknownFlags => {
+	console.error(`Unknown flag${unknownFlags.length > 1 ? 's' : ''}`);
+	console.error(unknownFlags.join('\n'));
```
sindresorhus, line 59: Same here. One `console` call.

weareoutman:
> In the future, don't force push. It makes it impossible for me to see what changed, and I'm forced to re-review all the code.

👌 Sorry for that, force pushed like my daily work and forgot about the review thing.

index.js
weareoutman: Fixed both.

kristian:
Just a suggestion, shouldn't the process rather fail with a exit code of 1 instead of 2? 2 is used when outputting the help and could be considered "not an error". Whereas if `allowUnknownFlags` is set to false, a unknown flag could be considered an error and thus should fail the process with an 1 insted of a 2?

sindresorhus:
Exit code 2 is the convention for invalid flags or other invalid usage: https://stackoverflow.com/questions/1101957/are-there-any-standard-exit-status-codes-in-linux/40484670#40484670

An exit code higher than 1 is AFAIK always an error.

sholladay:
> In the future, don't force push. It makes it impossible for me to see what changed, and I'm forced to re-review all the code.

@sindresorhus if you click on the phrase "force-pushed", it will actually show you the diff. Maybe Refined GitHub could make this more intuitive / discoverable.

sindresorhus:
Neat. I don't think it always did that though.

PR #181, opened 2021-04-25
Require Node.js 12 and move to ESM

### LINKED ISSUE #125: Can't run meow on the repl
Action: Run `require('meow')` from a node repl in node v12.7.0
Expected: no errors
Actual: Error `TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type object`

Caused by module.parent.filename being null at https://github.com/sindresorhus/meow/blob/master/index.js#L15

I made it work locally by using this code instead:
```javascript
let parentDir = ''
if (module.parent && module.parent.filename) {
  parentDir = path.dirname(module.parent.filename);
}
```
and it does work, but I'm not sure if that's a good solution since `parentDir` is used as `cwd`.

### LINKED ISSUE #157: Looks like `module.parent` is deprecated
I just saw it in the [Release Notes for Node v14.6.0](https://github.com/nodejs/node/releases/tag/v14.6.0). PR: https://github.com/nodejs/node/pull/32217

It is used here:

https://github.com/sindresorhus/meow/blob/fa2a3746acba0d68297685066ec665b88671db31/index.js#L13-L15

### LINKED ISSUE #176: Cannot read package information if using meow in ESM
`meow` is using `module.parent.filename` as `parentDir` to read package info:

```js
const parentDir = path.dirname(module.parent && module.parent.filename ? module.parent.filename : '.');
```

But in ESM, we cannot use [`__filename`](https://nodejs.org/dist/latest/docs/api/esm.html#esm_no_filename_or_dirname) anymore.

Do you have any plan move this package to ESM?

@sindresorhus And I wonder how you would solve this problem?

LitoMore:
Fixes #125
Resolves #157
Resolves #176


LitoMore:
Do I need to apply [this proposal](https://github.com/sindresorhus/meow/issues/157#issuecomment-662914347) in current PR?

sindresorhus:
> Do I need to apply this proposal #157 (comment) in current PR?

It doesn't have to be this PR, but it has to be done before what's in this PR can be released.

index.d.ts
```diff
@@ -309,7 +308,9 @@ const cli = meow(`
 foo(cli.input[0], cli.flags);
 ```
 */
+// eslint-disable-next-line no-redeclare
 declare function meow<Flags extends meow.AnyFlags>(helpMessage: string, options?: meow.Options<Flags>): meow.Result<Flags>;
+// eslint-disable-next-line no-redeclare
 declare function meow<Flags extends meow.AnyFlags>(options?: meow.Options<Flags>): meow.Result<Flags>;
```
sindresorhus, line 314: Just use `export default` on both of these. No need for declare.

sindresorhus:
The TS types are not correctly done. You need to remove the use of `namespace` and export the types.

test/is-required-flag.js
```diff
@@ -1,7 +1,9 @@
 import test from 'ava';
 import execa from 'execa';
-const path = require('path');
+import path from 'path';
```
sindresorhus, line 3:
```suggestion
import path from 'node:path';
```

For all the built-in modules. https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

sindresorhus (changes requested):

index.d.ts
```diff
@@ -309,7 +306,7 @@ const cli = meow(`
 foo(cli.input[0], cli.flags);
 ```
 */
-declare function meow<Flags extends meow.AnyFlags>(helpMessage: string, options?: meow.Options<Flags>): meow.Result<Flags>;
-declare function meow<Flags extends meow.AnyFlags>(options?: meow.Options<Flags>): meow.Result<Flags>;
 
-export = meow;
+declare function meow<Flags extends AnyFlags>(helpMessage: string | Options<Flags>, options?: Options<Flags>): Result<Flags>;
```
sindresorhus, line 310: This is not the same thing as previous. This allows specifying options twice.

sindresorhus (changes requested):

sindresorhus:
Please read https://github.com/sindresorhus/meow/pull/181#discussion_r619778517 more closely.

index.d.ts
```diff
@@ -1,25 +1,24 @@
...
 
-	/**
+/**
 	Callback function to determine if a flag is required during runtime.
 
 	@param flags - Contains the flags converted to camel-case excluding aliases.
 	@param input - Contains the non-flag arguments.
 
 	@returns True if the flag is required, otherwise false.
 	*/
```
sindresorhus, line 12: Incorrect indentation.

Applies everywhere. Please look at the file: https://github.com/LitoMore/meow/blob/esm/index.d.ts

index.d.ts
```diff
@@ -219,67 +218,65 @@ declare namespace meow {
...
 /**
 @param helpMessage - Shortcut for the `help` option.
 
 @example
 ```
 #!/usr/bin/env node
-'use strict';
-import meow = require('meow');
-import foo = require('.');
+import meow = from 'meow';
```
sindresorhus, line 278: Typo

index.d.ts
LitoMore: Do you mean use `export default` twice like this?

```typescript
export default function meow<Flags extends AnyFlags>(helpMessage: string, options?: Options<Flags>): Result<Flags>;
export default function meow<Flags extends AnyFlags>(options?: Options<Flags>): Result<Flags>;
```

index.d.ts
```diff
@@ -28,189 +27,189 @@ declare namespace meow {
...
+					return true;
 				}
+
+				return false;
 			}
 		}
-		```
+	}
+	```
 		*/
```
sindresorhus, line 64: Still incorrect indentation.

sindresorhus:
Looks good :)

PR #182, opened 2021-04-29
Allow user to pass in package URL for package reading

LitoMore:
Related to #176 

### Changes

- Add `importMeta` option for allowing users to pass in their package URL
- Apply `importMeta` option to all the test code

### Spec

- `importMeta` is required
- Its value should be `import.meta`

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -151,6 +151,7 @@ export interface Options<Flags extends AnyFlags> {
 			$ foo
 			🌈 unicorns✨🌈
 	`, {
+		packagePath: import.meta.url,
 		booleanDefault: undefined,
 		flags: {
 			rainbow: {
@@ -206,6 +207,13 @@ export interface Options<Flags extends AnyFlags> {
 	@default true
 	*/
 	readonly allowUnknownFlags?: boolean;
+
+	/**
+	Directory to start looking for the module package.json file.
+
+	@default process.cwd()
+	*/
+	readonly packagePath?: string;
 }
 
 type TypedFlag<Flag extends AnyFlag> =
--- a/index.d.ts
+++ b/index.d.ts
@@ -28,9 +28,7 @@ type AnyFlags = Record<string, AnyFlag>;
 
 export interface Options<Flags extends AnyFlags> {
 	/**
-	Your `import.meta` used for start looking for the module package.json file.
-
-	This option is required, its value must be `import.meta`.
+	Pass in `import.meta`. This is used to find the correct package.json file.
 	*/
 	readonly importMeta: ImportMeta;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -31,7 +31,7 @@ export interface Options<Flags extends AnyFlags> {
 	/**
 	Directory to start looking for the module package.json file.
 
-	This option is required, its value should be `import.meta`.
+	This option is required, its value must be `import.meta`.
 	*/
 	readonly importMeta: ImportMeta;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,7 +20,7 @@ export interface Flag<Type extends FlagType, Default> {
 	readonly isMultiple?: boolean;
 }
 
-type ImportMeta = typeof import.meta;
+type ImportMeta = Record<string, unknown> & {url: string};
 type StringFlag = Flag<'string', string>;
 type BooleanFlag = Flag<'boolean', boolean>;
 type NumberFlag = Flag<'number', number>;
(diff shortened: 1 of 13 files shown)


LitoMore:
We already have a `pkg` option for users to pass in their package object, what about let users read the package at their side, then pass it to our package?

index.js
```diff
@@ -97,18 +99,20 @@ const validateFlags = (flags, options) => {
...
 
-	const foundPackage = readPackageUpSync({
+	const foundPackage = readPackageSync({
+		cwd: options.packagePath ? dirname(fileURLToPath(options.packagePath)) : process.cwd(),
 		normalize: false
 	});
 
 	options = {
-		pkg: foundPackage ? foundPackage.packageJson : {},
+		packagePath: process.cwd(),
```
LitoMore, line 114: The default value is `process.cwd()`, but it's useless. Do we need to set it as a required option?

sindresorhus:
I think we should accept `import.meta`, not `import.meta.url`. `import.meta` might get useful properties we might want to use in the future. It's more future proof this way. We can just call the option `importMeta` and make it required.

index.js
```diff
@@ -1,11 +1,13 @@
...
+import {fileURLToPath} from 'node:url';
 import buildParserOptions from 'minimist-options';
 import parseArguments from 'yargs-parser';
 import camelCaseKeys from 'camelcase-keys';
 import decamelize from 'decamelize';
 import decamelizeKeys from 'decamelize-keys';
 import trimNewlines from 'trim-newlines';
 import redent from 'redent';
-import {readPackageUpSync} from 'read-pkg-up';
+import {readPackageSync} from 'read-pkg';
```
sindresorhus, line 10: The `cli.js` is not guaranteed to be in the same directory as package.json.

index.d.ts
```diff
@@ -20,13 +20,21 @@ export interface Flag<Type extends FlagType, Default> {
...
 type NumberFlag = Flag<'number', number>;
 type AnyFlag = StringFlag | BooleanFlag | NumberFlag;
 type AnyFlags = Record<string, AnyFlag>;
 
 export interface Options<Flags extends AnyFlags> {
+	/**
+	Directory to start looking for the module package.json file.
+
+	This option is required, its value must be `import.meta`.
+	*/
```
sindresorhus, line 35: Docs are not correct.

index.d.ts
```diff
@@ -20,13 +20,21 @@ export interface Flag<Type extends FlagType, Default> {
 	readonly isMultiple?: boolean;
 }
 
+type ImportMeta = {url: string};
```
sindresorhus, line 23:
```suggestion
type ImportMeta = typeof import.meta;
```

?

LitoMore: It looks like it's not correct grammar.

LitoMore: How about use `Record<string, unknown> & {url: string}` instead?

LitoMore: We could use `ImportMeta` directly.

The type definitions is here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/globals.d.ts#L29

index.js
```diff
@@ -97,13 +99,18 @@ const validateFlags = (flags, options) => {
...
 
-const meow = (helpText, options) => {
+const meow = (helpText, options = {}) => {
 	if (typeof helpText !== 'string') {
 		options = helpText;
 		helpText = '';
 	}
 
+	if (!(options.importMeta && options.importMeta.url)) {
+		throw new TypeError('`importMeta` options is required, its value must be `import.meta`');
```
sindresorhus, line 109:
```suggestion
		throw new TypeError('The `importMeta` option is required. Its value must be `import.meta`.');
```

index.d.ts
```diff
@@ -27,6 +27,13 @@ type AnyFlag = StringFlag | BooleanFlag | NumberFlag;
 type AnyFlags = Record<string, AnyFlag>;
 
 export interface Options<Flags extends AnyFlags> {
+	/**
+	Your `import.meta` used for start looking for the module package.json file.
+
+	Set it to `import.meta` to find the module package.json file.
```
sindresorhus, line 33: These two sentence say almost the same thing.

readme.md
```diff
@@ -126,6 +86,14 @@ Shortcut for the `help` option.
 
 Type: `object`
 
+##### importMeta
+
+Type: `object`
+
+Your `import.meta` used for start looking for the module package.json file.
+
+This option is required, its value must be `import.meta`.
```
sindresorhus, line 95: Readme and index.d.ts should be in sync.

sindresorhus (changes requested):

index.d.ts
```diff
@@ -27,6 +27,13 @@ type AnyFlag = StringFlag | BooleanFlag | NumberFlag;
 type AnyFlags = Record<string, AnyFlag>;
 
 export interface Options<Flags extends AnyFlags> {
+	/**
+	Your `import.meta` used for start looking for the module package.json file.
+
+	This option is required, its value must be `import.meta`.
```
sindresorhus, line 33:
```suggestion
	Pass in `import.meta`. This is used to find the correct package.json file.
```

PR #183, opened 2021-05-22
Add choices option

### LINKED ISSUE #171: Add `choices` option
I want to define a type like this:
```js
      aFlag: {
        type: "global" | "local",
        default: "global",
      },
```

How this possible? or if that is not possible. could you implement it?

toshihidetagami:
Resolves #171 .

- string, number can be used for choices array. Boolean value(true, false) can also be used though it does not make sense.
- Flag from choices is required if isRequired is true, otherwise not required.
- If isMultiple is true, user can pass multiple arguments from choices.

Let me know if there are any problems with spec/behavior.

### DIFF
--- a/index.js
+++ b/index.js
@@ -58,6 +58,37 @@ const validateOptions = ({flags}) => {
 	}
 };
 
+const validateChoices = (flags, receivedFlags) => {
+	const errors = [];
+
+	for (const [flagKey, flagValue] of Object.entries(flags)) {
+		const {choices, isRequired} = flagValue;
+
+		if (!choices) {
+			return;
+		}
+
+		if (!Array.isArray(choices)) {
+			throw new TypeError('Choices should be array');
+		}
+
+		const receivedInput = receivedFlags[flagKey];
+		if (receivedInput === undefined && !isRequired) {
+			return;
+		}
+
+		const receivedInputArray = Array.isArray(receivedInput) ? receivedInput : [receivedInput];
+		const includesAll = receivedInputArray.every(i => choices.includes(i));
+		if (!includesAll) {
+			errors.push(`Choices: [${choices.join(', ')}] Received: [${receivedInputArray.join(', ')}]`);
+		}
+	}
+
+	if (errors.length > 0) {
+		throw new Error(`Choices mismatch: ${errors.join(', ')}`);
+	}
+};
+
 const reportUnknownFlags = unknownFlags => {
 	console.error([
 		`Unknown flag${unknownFlags.length > 1 ? 's' : ''}`,
@@ -209,6 +240,7 @@ const meow = (helpText, options = {}) => {
 	const unnormalizedFlags = {...flags};
 
 	validateFlags(flags, options);
+	validateChoices(options.flags, flags);
 
 	for (const flagValue of Object.values(options.flags)) {
 		delete flags[flagValue.alias];
--- a/test/test.js
+++ b/test/test.js
@@ -579,6 +579,119 @@ test('isMultiple - handles multi-word flag name', t => {
 	});
 });
 
+test('choices - success case', t => {
+	const cli = meow({
+		importMeta,
+		argv: ['--animal', 'cat', '--number=2.2'],
+		flags: {
+			animal: {
+				choices: ['dog', 'cat', 'unicorn']
+			},
+			number: {
+				choices: [1.1, 2.2, 3.3],
+				type: 'number'
+			}
+		}
+	});
+
+	t.is(cli.flags.animal, 'cat');
+	t.is(cli.flags.number, 2.2);
+});
+
+test('choices - throws if input does not match choices', t => {
+	t.throws(() => {
+		meow({
+			importMeta,
+			argv: ['--animal', 'rainbow', '--number', 5],
+			flags: {
+				animal: {
+					choices: ['dog', 'cat', 'unicorn']
+				},
+				number: {
+					choices: [1, 2, 3]
+				}
+			}
+		});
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -58,6 +58,37 @@ const validateOptions = ({flags}) => {
...
+			throw new TypeError('Choices should be array');
+		}
+
+		const receivedInput = receivedFlags[flagKey];
+		if (receivedInput === undefined && !isRequired) {
+			return;
+		}
+
+		const receivedInputArray = Array.isArray(receivedInput) ? receivedInput : [receivedInput];
+		const includesAll = receivedInputArray.every(i => choices.includes(i));
```
sindresorhus, line 81:
```suggestion
		const includesAll = receivedInputArray.every(index => choices.includes(index));
```

test/test.js
```diff
@@ -579,6 +579,119 @@ test('isMultiple - handles multi-word flag name', t => {
...
+			argv: ['--animal=dog', '--animal=rabbit'],
+			flags: {
+				animal: {
+					type: 'string',
+					isMultiple: true,
+					choices: ['dog', 'cat', 'unicorn']
+				}
+			}
+		});
+	}, {message: 'Choices mismatch: Choices: [dog, cat, unicorn] Received: [dog, rabbit]'});
```
sindresorhus, line 692: I think the message could be more succinct:

```suggestion
	}, {message: 'Unknown value `rabbit`. Value must be one of: dog, cat, unicorn'});
```

sindresorhus (changes requested):

sindresorhus:
Thanks for working on this. You also need to update index.d.ts.

sindresorhus:
@toshihidetagami Friendly bump :)

toshihidetagami:
@sindresorhus Thanks for the review and sorry that it took so long. I added commits to this PR.

index.d.ts
```diff
@@ -18,6 +18,7 @@ export interface Flag<Type extends FlagType, Default, IsMultiple = false> {
 	readonly default?: Default;
 	readonly isRequired?: boolean | IsRequiredPredicate;
 	readonly isMultiple?: IsMultiple;
+	readonly choices?: any[];
```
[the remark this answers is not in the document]
toshihidetagami: I made the following change.

### Specification

If type is specified, choices should be an array of this type.
If type is not specified, choices can be any 3 types of an array but can’t contain different types. I considered accepting heterogeneous arrays if types are not specified, but choose homogeneous arrays for the following reasons. Shamefully, the biggest reason is my inability to implement, let me know if this will be the problem.

- It seems to require multiple changes in the type definition around flags.
- I think there are few valid cases to use heterogeneous arrays.
- isMultiple also accepts only homogeneous arrays.

### About Code

**index.d.ts**
I renamed `Default` to `Type` because it will also be used for choices so it needs to be a more general name. And I renamed the original `Type` to `LiteralType`.

index.d.ts
```diff
@@ -12,12 +12,13 @@ Callback function to determine if a flag is required during runtime.
...
-export interface Flag<Type extends FlagType, Default, IsMultiple = false> {
-	readonly type?: Type;
+export interface Flag<LiteralType extends FlagType, Type, IsMultiple = false> {
+	readonly type?: LiteralType;
 	readonly alias?: string;
-	readonly default?: Default;
+	readonly default?: Type;
 	readonly isRequired?: boolean | IsRequiredPredicate;
 	readonly isMultiple?: IsMultiple;
+	readonly choices?: Type extends any[] ? Type : Type[];
```
sindresorhus, line 21:
```suggestion
	readonly choices?: Type extends unknown[] ? Type : Type[];
```

test/test.js
```diff
@@ -579,6 +579,135 @@ test('isMultiple - handles multi-word flag name', t => {
...
+	const cli = meow({
+		importMeta,
+		argv: ['--animal', 'cat', '--number=2.2'],
+		flags: {
+			animal: {
+				choices: ['dog', 'cat', 'unicorn'],
+			},
+			number: {
+				choices: [1.1, 2.2, 3.3],
+				type: 'number',
```
sindresorhus, line 592: Nit: `type` should be first.

sindresorhus (changes requested):

index.d.ts
toshihidetagami: thanks for pointing out this

index.js
```diff
@@ -58,6 +58,55 @@ const validateOptions = ({flags}) => {
...
+	for (const [flagKey, flagValue] of Object.entries(flags)) {
+		const receivedInput = receivedFlags[flagKey];
+		const errorMessage = validateChoicesByFlag(flagKey, flagValue, receivedInput);
+		if (errorMessage) {
+			errors.push(errorMessage);
+		}
+	}
+
+	if (errors.length > 0) {
+		throw new Error(`${errors.join('. ')}`);
```
LitoMore, line 106: Please add test for case of multiple errors.

LitoMore (changes requested):

index.js
LitoMore: @sindresorhus The output from here will be:

```
Unknown values: `dog, rabbit`. Value must be one of: cat, unicorn. Unknown values: `dog, rabbit`. Value must be one of: cat, unicorn. Unknown values: `dog, rabbit`. Value must be one of: cat, unicorn
```

Is this acceptable for you?

LitoMore: We are using `new Error()` for throwing error to developers. But this error message is for users. We should use `console.error` instead.

sindresorhus:
Bump :)

PR #185, opened 2021-06-08
Upgrade dependencies

resynthesize:
trim-newlines 4.0.0 has a high severity CVE vulnerability noted here: https://www.npmjs.com/advisories/1753 - this PR updates to the version mentioned in the CVE.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
 		"normalize-package-data": "^3.0.2",
 		"read-pkg-up": "^8.0.0",
 		"redent": "^4.0.0",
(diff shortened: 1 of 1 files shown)


sindresorhus:
To be clear, it doesn't affect `meow` at all.

pitgrap:
Afaik "^" with semver already covered this update and I don't see any lock file in this repository. But anyways thanks you for the update and release!

voxpelli:
@pitgrap This way users of `meow` will get the subdependency update for free when they update `meow`, else they need to do a lock-file update as well

Issue #189, opened 2021-07-15
Typescript definition for default does not support arrays


RobertSasak:
Property _default_ expects _string_, _boolean_, _number_. However when _isMultiple_ is _true_ it should also accepts arrays. 

Here is a minimal example of the issue:
```ts
import meow from 'meow'

meow(`Usage`, {
    importMeta: import.meta,
    flags: {
        unicorn: {
            type: 'string',
            alias: 'u',
            default: ['rainbow', 'cat'], // <- Type issue
            isMultiple: true,
        },
    },
})
```
Actual error message:
```
Type 'string[]' is not assignable to type 'string | number | boolean | undefined'.
  Type 'string[]' is not assignable to type 'string'.ts(2322)
index.d.ts(18, 11): The expected type comes from property 'default' which is declared here on type 'AnyFlag'
(property) Flag<"string", string>.default?: string | undefined
```

PR #190, opened 2021-07-15
Allow `default` property of `Flag` types to accept arrays

RobertSasak:
Fixes: https://github.com/sindresorhus/meow/issues/189


sindresorhus:
Can you add a type test?

And when `isMultiple: true`, it should only be an array.

RobertSasak:
Thank you for teaching me _tsd_

sindresorhus:
Thanks :)

PR #194, opened 2021-07-21
Fix failure with `isMultiple` when `isRequired` function returns `false`

### LINKED ISSUE #193: isMultiple is still required when isRequired fn returns false
Example:
``` js
const cli = meow({
	importMeta: import.meta,
	description: "Custom description",
	help: `
		Usage
		  foo <input>
  `,
	flags: {
		test: {
			type: "number",
			alias: "t",
			isRequired: () => false,
			isMultiple: true,
		},
	},
});

console.log(cli.flags.test);
```

should return `[]`

but instead returns

``` bash
Missing required flag
        --test, -t
```

softwarewright:
Fixes #193

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,7 +26,7 @@ const isFlagMissing = (flagName, definedFlags, receivedFlags, input) => {
 		return isFlagRequired;
 	}
 
-	return flag.isMultiple && receivedFlags[flagName].length === 0;
+	return flag.isMultiple && receivedFlags[flagName].length === 0 && isFlagRequired;
 };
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
Thanks for the pull request. Always ensure running `npm test` locally passes before pushing commits.

softwarewright:
@sindresorhus Ah sorry assumed it was a local issue. I'll look at why the build is failing.

softwarewright:
@sindresorhus I fixed the lint errors, the other `typescript` issue seems to be with the `xo` package, but I think you are already aware of this: https://github.com/xojs/xo/issues/555.

PR #201, opened 2021-09-30
update xo

ivanoats:
resolve security update


sindresorhus:
XO is a dev dependency, so it shouldn't cause any security notices when installing `meow`. I'm also not interested in a lock file.

PR #206, opened 2022-03-24
Upgrade xo (minor).

millette:


index.d.ts
```diff
@@ -197,7 +197,7 @@ export interface Options<Flags extends AnyFlags> {
 	//}
 	```
 	*/
-	readonly booleanDefault?: boolean | null | undefined;
+	readonly booleanDefault?: boolean | undefined;
```
sindresorhus, line 200: You cannot just remove this.

test/fixtures/fixture.js
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
+
+import process from 'node:process';
```
sindresorhus, line 3:
```suggestion
import process from 'node:process';
```

PR #207, opened 2022-04-21
Bump dependencies

LitoMore:
I kept the `ava` at version 3. For `ava@4` our test will fail. I haven't looked into it yet. But I can give a check with another PR.

I made denpendencies updates and GitHub Actions into one PR. Please let me know if you want to separate them into two PRs.

Resolves #199 
Closes #206

### Changes

- Bump dependencies to latest
- Add Node.js 14 and 18 to CI matrix

PR #208, opened 2022-04-21
Use Node.js 18

LitoMore:
### Changes

- Add Node.js 14 and 18 to matrix.


LitoMore:
Mergin #207 may fix the failing CI for Node.js 14.

PR #209, opened 2022-05-03
feat: introduce `preferNestedFlags` option

antongolub:
[Real-world issue](https://github.com/qiwi-forks/docma/runs/6279032275?check_suite_focus=true#step:4:40): I got stuck in a corner case when I couldn't parametrize CLI inside another CLI:

```shell
npx -p foo -p bar bar-cli -- --opts-for='bar-cli'

Too many non-option arguments: got 2, maximum of 0
Error: Process completed with exit code 1.
```

I suggest introducing `preferNestedFlags` option to workaround the limitation:
```js
const argvStart = options.preferNestedFlags && process.argv.includes('--') ? process.argv.indexOf('--') + 1 : 2;

options = {
    pkg: foundPackage ? foundPackage.packageJson : {},
    argv: process.argv.slice(argvStart),
    ...
};
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -292,6 +292,13 @@ Default: `true`
 
 Whether to allow unknown flags or not.
 
+##### preferNestedFlags
+
+Type `boolean`\
+Default: `false`
+
+This option makes sense if your CLI may be invoked inside another one. For example, `npx -p foo -p bar bar-cli -- --opts-for='bar-cli'`.
+
 ## Promises
 
 Meow will make unhandled rejected promises [fail hard](https://github.com/sindresorhus/hard-rejection) instead of the default silent fail. Meaning you don't have to manually `.catch()` promises used in your CLI.
--- a/index.d.ts
+++ b/index.d.ts
@@ -213,6 +213,13 @@ export interface Options<Flags extends AnyFlags> {
 	@default true
 	*/
 	readonly allowUnknownFlags?: boolean;
+
+	/**
+	Prefer nested flags to be parsed first.
+
+	@default false
+	 */
+	readonly preferNestedFlags?: boolean;
 }
 
 type TypedFlag<Flag extends AnyFlag> =
--- a/index.js
+++ b/index.js
@@ -116,9 +116,11 @@ const meow = (helpText, options = {}) => {
 		normalize: false,
 	});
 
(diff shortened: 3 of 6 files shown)


antongolub:
@sindresorhus, what do you think of that?

sindresorhus:
I would recommend trying to find a way to solve this without introducing a new flag. Maybe open an issue on https://github.com/yargs/yargs-parser (the parser `meow` uses) with your problem statement.

antongolub:
There's no need for yargs-* patching. I just thought a flag would be a bit more convenient than:
```js
const cli = meow({
	importMeta: import.meta,
	description: 'Custom description',
	help: `
		Usage
		  foo <input>
  `,
	// preferNestedFlags: true,
	flags: {
		foo: {
			type: 'string',
		},
	},
	get argv() {
		const argvStart = process.argv.includes('--') ? process.argv.indexOf('--') + 1 : 2
		return process.argv.slice(argvStart)
	}
});
```

sindresorhus:
If a flag actually makes sense for this, it makes more sense to add it to the actual argument parser.

readme.md
```diff
@@ -292,6 +292,13 @@ Default: `true`
 
 Whether to allow unknown flags or not.
 
+##### preferNestedFlags
+
+Type `boolean`\
+Default: `false`
+
+This option makes sense if your CLI may be invoked inside another one. For example, `npx -p foo -p bar bar-cli -- --opts-for='bar-cli'`.
```
sindresorhus, line 300: **Any** CLI may be invoked inside another one. I really don't think this is the best way to resolve your problem.

antongolub: My issue may be related to npx: seems like it loads my CLI in its own process. I'll check the `process.argv[1]` value.

antongolub:
1.  everything is fine with npx. Side effect was produced by our own cli composer
2. `-- --foo='bar'`  is more of an aesthetic wish. `npx -p pkg1 -p pkg2 some-cli --foo='bar'` works fine
3. `get argv() {
  const argvStart = process.argv.includes('--') ? process.argv.indexOf('--') + 1 : 2
  return process.argv.slice(argvStart)
}` completely covers my issue

PR #213, opened 2022-06-21
Set showHelp return type to never

ryanrhee:
Since `showHelp()` calls into `process.exit()`, they should use the same return type.

The return type of `process.exit()` is `never`, not `void`. This tells the typescript compiler that code branches that have `showHelp()` terminate, and that any code written as executing after `showHelp()` should be considered dead code (similar to code after a `return` statement).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -268,7 +268,7 @@ export interface Result<Flags extends AnyFlags> {
 
 	@param exitCode - The exit code to use. Default: `2`.
 	*/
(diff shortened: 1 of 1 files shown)


ryanrhee:
As a workaround:

```ts
  const showHelp = () => {
    cli.showHelp()
    throw new Error("This is unreachable code")
  }
  
  // type of showHelp() is () => never
```

PR #225, opened 2023-03-18
Rename `alias` to `shortFlag`

### LINKED ISSUE #109: Rename `alias` option to `shortFlag`
I never liked the `alias` wording, which was inherited from `minimist`. I think it would be better to have a more explicit and semantic name called `shortFlag`.

We could then also add a `aliases` option where users can specify actual aliases and not just limited to one:

```js
{
	flags: {
		unicorn: {
			type: 'string',
			default: 'love',
			shortFlag: 'u',
			aliases: [
				'--rainbow',
				'-r'
			]
		}
	}
}
```

Could use some feedback on this proposal.

tommy-mitchell:
Closes #109.

`flag.alias` is now `flag.shortFlag` for users:

```js
const cli = meow(`
	Usage
	  $ foo <input>

	Options
	  --rainbow, -r  Include a rainbow

	Examples
	  $ foo unicorns --rainbow
	  🌈 unicorns 🌈
`, {
	importMeta: import.meta,
	flags: {
		rainbow: {
			type: 'boolean',
			shortFlag: 'r'
		}
	}
});
```

An extra check was added to `buildParserFlags()` to ensure compatibility with `minimist-options`:

```js
// index.js

const buildParserFlags = ({flags, booleanDefault}) => {
	const parserFlags = {};

	for (const [flagKey, flagValue] of Object.entries(flags)) {
		const flag = {...flagValue};

		// `buildParserOptions` expects `flag.alias`
		if (flag.shortFlag) {
			flag.alias = flag.shortFlag;
			delete flag.shortFlag;
		}

		// ...
	}

	return parserFlags;
}
```

### DIFF
--- a/estest/index.js
+++ b/estest/index.js
@@ -17,7 +17,7 @@ meow(
 		flags: {
 			rainbow: {
 				type: 'boolean',
-				alias: 'r',
+				shortFlag: 'r',
 			},
 		},
 	},
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,7 +17,7 @@ export type IsRequiredPredicate = (flags: Readonly<AnyFlags>, input: readonly st
 
 export type Flag<Type extends FlagType, Default, IsMultiple = false> = {
 	readonly type?: Type;
-	readonly alias?: string;
+	readonly shortFlag?: string;
 	readonly default?: Default;
 	readonly isRequired?: boolean | IsRequiredPredicate;
 	readonly isMultiple?: IsMultiple;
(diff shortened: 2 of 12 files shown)


tommy-mitchell:
Should `flag.aliases` be a part of this PR as well, or a separate one?

sindresorhus:
> Should flag.aliases be a part of this PR as well, or a separate one?

Separate

sindresorhus:
Can you make it throw a user-friendly error when `alias` option is used?

tommy-mitchell:
> Can you make it throw a user-friendly error when `alias` option is used?

```js
meow({
	importMeta,
	flags: {
		foo: {
			type: 'string',
			alias: 'f',
		},
		bar: {
			type: 'string',
			alias: 'b',
		},
		baz: {
			type: 'string',
			shortFlag: 'z',
		},
	},
});
//=> 'The option `alias` has been renamed to `shortFlag`. The following flags need to be updated: `foo`, `bar`'
```

sindresorhus:
Perfect 👍

PR #228, opened 2023-03-21
Add `choices` option

### LINKED ISSUE #171: Add `choices` option

tommy-mitchell:
Closes #171. Continues from #183. Improves the error messages, adds tests cases, and aggregates option validation error messages.

Currently the choices array isn't validated to be the same type as the flag.

There were a couple of days in between starting and creating this PR, so I may have missed something.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -129,7 +129,7 @@ test('flag declared in kebab-case is an error', t => {
 			flags: {'kebab-case': 'boolean', test: 'boolean', 'another-one': 'boolean'},
 		});
 	});
-	t.is(error.message, 'Flag keys may not contain \'-\': kebab-case, another-one');
+	t.is(error.message, 'Flag keys may not contain \'-\'. Invalid flags: `kebab-case`, `another-one`');
 });
 
 test('type inference', t => {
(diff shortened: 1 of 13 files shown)


test/test.js
```diff
@@ -683,6 +684,200 @@ test('aliases - unnormalized flags', t => {
...
+				},
+			},
+		});
+	}, {message: stripIndent`
+		Unknown value for flag \`animal\`: \`dog\`. Value must be one of: [cat, unicorn]
+		Unknown value for flag \`plant\`: \`succulent\`. Value must be one of: [tree, flower]
+	`});
+});
+
+test.failing('choices - choices must be of the same type', t => {
```
sindresorhus, line 842:
```suggestion
test('choices - choices must be of the same type', t => {
```

test/test.js
```diff
@@ -683,6 +684,200 @@ test('aliases - unnormalized flags', t => {
...
+					choices: ['cat', 'unicorn'],
+				},
+				plant: {
+					type: 'string',
+					choices: ['tree', 'flower'],
+				},
+			},
+		});
+	}, {message: stripIndent`
+		Unknown value for flag \`animal\`: \`dog\`. Value must be one of: [cat, unicorn]
```
sindresorhus, line 837:
```suggestion
		Unknown value for flag \`--animal\`: \`dog\`. Value must be one of: [cat, unicorn]
```

sindresorhus (changes requested):

tommy-mitchell:
Updated with kebab-cased flags in errors and `choices` type validation.

PR #238, opened 2023-05-03
Fix flag.default values validation

### LINKED ISSUE #239: flag.default values validation is faulty
Their is an issue in the validation of the default values .

Example:
```
meow({
	importMeta,
	flags: {
		string: {
			type: 'string',
			choices: ['dog', 'cat', 'unicorn'],
			default: 'unicorn',
		},
	},
});
```
should not throw an error, since 'unicorn' exists within the option `choices`.

But instead, it throws the following error:
```
Each value of the option `default` must exist within the option `choices`. Invalid flags: `--string`.
```

istandre:
Fixes #239
This fixes the validation of the default values.

Example:
```
meow({
	importMeta,
	flags: {
		string: {
			type: 'string',
			choices: ['dog', 'cat', 'unicorn'],
			default: 'unicorn',
		},
	},
});
```
should not throw an error, since 'unicorn' exists within the option `choices`.

But instead, it throws the following error:
```
Each value of the option `default` must exist within the option `choices`. Invalid flags: `--string`.
```

### DIFF
--- a/source/options.js
+++ b/source/options.js
@@ -27,7 +27,7 @@ const validateOptions = options => {
 				},
 			},
 			defaultNotInChoices: {
-				filter: ([, flag]) => flag.default && Array.isArray(flag.choices) && [flag.default].flat().every(value => flag.choices.includes(value)),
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks :)

PR #242, opened 2023-08-01
Use Rollup for bundling dependencies

### LINKED ISSUE #67: Bundle dependencies

tommy-mitchell:
Closes #67. Bumps dev dependencies and adds an `_actualDependencies` field.

Builds on `prepare` and outputs a bundled `dependencies.js` and `licenses.md` to the `build` folder:

```sh
\__ build/
	\__ licenses.md # License of each bundled dependency
	\__ dependencies.js # Vendor bundle
	\__ index.js
	\__ options.js
	\__ parser.js
	\__ utils.js
	\__ validate.js
\__ index.d.ts
\__ license
\__ package.json
\__ readme.md
```


sindresorhus:
I don't want to minify the code. Debugging is important. It also doesn't save much anyway.

sindresorhus:
> How should dependencies in package.json be specified?

They need to be devDependencies, but it would be nice to somehow indicate which are used in production. We could maybe create a custom field like in package.json: `{"_actualDependencies": ["..."]}`.

package.json
```diff
@@ -13,16 +13,19 @@
 	"type": "module",
 	"exports": {
 		"types": "./index.d.ts",
-		"default": "./source/index.js"
+		"default": "./dist/index.js"
```
sindresorhus, line 16: I kinda want to go away from abbreviations.

Which one do you like the most?

- distribution
- output
- release
- production

tommy-mitchell: Maybe `build`? Of those, I like `output`.

tommy-mitchell:
> They need to be devDependencies, but it would be nice to somehow indicate which are used in production. We could maybe create a custom field like in package.json: `{"_actualDependencies": ["..."]}`.

Would `bundleDependencies` work?

sindresorhus:
> Would bundleDependencies work?

I don't think so. It would bundle the dependencies. So we would still have double up of the code (both rolled-up and npm bundled).

sindresorhus:
I think it's easier to simply bundle everything. If we do a separate dependencies file, we need to change all the import statements.

tommy-mitchell:
Whoops, you're right, was thinking a little too fast there.

tommy-mitchell:
I'm not sure when the build script should be run. Pros and cons:

- `prepack`:
	- [recommended](https://yarnpkg.com/advanced/lifecycle-scripts) by `yarn`
	- runs when packages are installed from `git`
	- also runs when `np` does checks (e.g. `npm pack --dry-run --json`)
		- [seems like](https://docs.npmjs.com/cli/v9/commands/npm-pack#dry-run) it shouldn't, but I'm not sure
- `prepublishOnly`:
	- only runs on `npm publish`
	- doesn't run until needed, but `git` consumers can't consume `meow`

LitoMore:
@tommy-mitchell FYI, `Got` and `Ky` are using `prepare`.

- https://github.com/sindresorhus/got/blob/main/package.json
- https://github.com/sindresorhus/ky/blob/main/package.json

sindresorhus:
Either `prepack` or `prepare`. I generally prefer `prepare` as it also executes when running `npm install` while developing.

tommy-mitchell:
Changes:

* Added an `_actualDependencies` field to `package.json`
* Changed output directory to `build` instead of `dist`
* Changed build script to use `prepare` instead of `prepack`

With Rollup, I was able to separate the dependencies into a separate file and have the imports be updated:

```
\__ build/
	\__ licenses.md # License of each bundled dependency
	\__ dependencies.js # Vendor bundle
	\__ index.js
	\__ options.js
	\__ parser.js
	\__ utils.js
	\__ validate.js
\__ index.d.ts
\__ license
\__ package.json
\__ readme.md
```

package.json
```diff
@@ -13,16 +13,19 @@
...
-		"default": "./source/index.js"
+		"default": "./build/index.js"
 	},
+	"sideEffects": false,
 	"engines": {
 		"node": ">=16.10"
 	},
 	"scripts": {
+		"prepare": "npm run build",
+		"build": "rollup -c",
```
sindresorhus, line 24: Use the long flag.

rollup.config.js
```diff
@@ -0,0 +1,50 @@
+import json from '@rollup/plugin-json';
+import commonjs from '@rollup/plugin-commonjs';
+import {nodeResolve} from '@rollup/plugin-node-resolve';
+import license from 'rollup-plugin-license';
+import strip from 'strip-comments';
+import {defineConfig} from 'rollup';
+
+const outDir = 'build';
```
sindresorhus, line 8:
```suggestion
const outputDirectory = 'build';
```

sindresorhus (changes requested):

sindresorhus:
Looks pretty good

sindresorhus:
Have you manually tested it with a CLI to make sure `meow` still works when published? You can run `npm pack` and then decompress it in a CLI project.

tommy-mitchell:
> Have you manually tested it

I had before, but I went ahead and added a couple of tests using the built output to verify going forward.

tommy-mitchell:
> I guess it doesn't hurt, but kinda weird that each generated file includes

Yeah I'm not sure what's up with that, I need to do some research.

test/fixtures/build.js
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env node
```
LitoMore, line 1: Is this fixture in use? You may need to `chmod +x` this file to executable.

tommy-mitchell: It is and I did, I believe :)

LitoMore: My bad. It's already executable. The GitHub changes didn't show its file mode.

test/build.js
```diff
@@ -0,0 +1,40 @@
+/* eslint-disable ava/no-ignored-test-files */
```
LitoMore, line 1: Could you check with `npx xo` if this error still exists?

I remember this error only appears on the editor somehow.

LitoMore: I see. You've added a new line `!test/build.js` to the `ava.files`.

Please ignore my comment if you are intentionally ignoring this file.

tommy-mitchell: Yeah, this test runs separate from the others. I guess the `test` script could be set up to build before running, so they could be together.

tommy-mitchell:
Managed to fix the empty Node builtin imports. I think this should be good to go.

Grohden:
@sindresorhus @tommy-mitchell 
One question about this PR: How does it work with security checks and resolution bumps? Since dependencies are now bundled and not declared in the package.json, security issues with them might not be reported anymore by bots... we will also not be able to properly check if a dependency is updated

In the old version:
```
➜  project git:(main) ✗ yarn why semver
└─ normalize-package-data@npm:2.5.0
   └─ semver@npm:5.7.2 (via npm:2 || 3 || 4 || 5)
➜  project git:(main) ✗ yarn why normalize-package-data
├─ meow@npm:6.1.1
│  └─ normalize-package-data@npm:2.5.0 (via npm:^2.5.0)
│
└─ read-pkg@npm:5.2.0
   └─ normalize-package-data@npm:2.5.0 (via npm:^2.5.0)
➜  project git:(main) ✗ 
```

In the new bundled deps version (12.1.1):
```
➜  project git:(main) ✗ yarn why normalize-package-data
└─ read-pkg@npm:5.2.0
   └─ normalize-package-data@npm:2.5.0 (via npm:^2.5.0)
```

No report, although `normalize-package-data` it's being used by `meow`

Also, how does security issues are auto checked in this repo?

sindresorhus:
> not be reported anymore by bots

I would argue that's an upside. 99% of all npm audits are not actually vulnerabilities (the `semver` "vulnerability" is a good example of one) or not applicative to `meow`. So this reduces the noise.

https://overreacted.io/npm-audit-broken-by-design/

Grohden:
@sindresorhus I agree that audit sucks. Big red "critical" text scares security company monke

image


But as Dan says "Inlining dependencies kind of goes against the whole point of npm", and I feel like we kinda lose any analysis (you might not only use it for security) tools using that approach:

* I can't know for sure this dep is in my codebase (if its a real issue)
* I can't report it in bundle analysis (it's gonna report meow as a bigger dep)
* I can't use resolutions field to force dependency version (this is somehow a win.. cuz its a nightmare to remember potential issues with the resolutions)
* More potential to duplicate code (ties back to what Dan said)

But in the end it's your decision & opinion as a lib owner, as long as you're aware I think its ok

sindresorhus:
I generally agree that inlining is bad, but in this specific case, the upsides outweight the downsides.

fregante:
I'm on the fence with this change… but also I think it would be great to see it on `ora`, which seems to be super small but in reality: 

- https://npmgraph.js.org/?q=ora@7.0.1
- https://packagephobia.com/result?p=ora (more than [meow](https://packagephobia.com/result?p=meow) itself, now)

PR #246, opened 2023-09-18
fix: use default types location to comply legacy tsc moduleResolution

antongolub:
12.1.1 brings a new types location, which differs from the default `./index.d.ts`, so now it breaks transpilation for many _legacy_ tsconfigs. Let's keep it compatible.

https://github.com/microsoft/TypeScript/issues/51862

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,10 +11,8 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
-	"exports": {
-		"types": "./build/index.d.ts",
-		"default": "./build/index.js"
-	},
+	"types": "./build/index.d.ts",
+	"exports": "./build/index.js",
 	"sideEffects": false,
 	"engines": {
 		"node": ">=16.10"


sindresorhus:
No thanks. This is an ESM package and you are expected to use `moduleResolution: node16`.

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

antongolub:
Ok, how about deprecating 12.1.1 and rolling it up as 13.0.0? 12.1.1 — brought us unexpected breaking change.

antongolub:
```ts
// meow.d.ts
declare module "meow" {
  import type * as meow from 'meow/build/index.d.ts'
  export = meow
}
```

sindresorhus:
The change that defined `exports` in package.json was released with v11, in a major version, a long time ago. The change in 12.1.1 was techincally not breaking. It only breaks because you were using the package even though your project is not ESM.

sindresorhus:
I suggest pinning the version until you move to ESM.

antongolub:
> you were using the package even though your project is not ESM
I'm afraid, our case is a bit exotic: we use ESM ("exports, type: module" in pkg.json,  "module": "esnext" in tsconfig), but we do not inject `.js` extensions to `ts` files. Instead, we use [tsc-esm-fix](https://github.com/antongolub/tsc-esm-fix) for post-build patching.

Issue #233, opened 2023-03-28
Refactor


tommy-mitchell:
Per the discussion in https://github.com/sindresorhus/meow/pull/230#issuecomment-1483644249, this is a tracker issue for a general refactor:

- [ ] Target Node 14 and update dependencies
- [ ] Separate `index.js` into separate source files
	- Separate tests into smaller files/fixtures, maybe consolidate via macros
	- Add doc comments to functions
	- Not directly related, but adding subcommands (#69) benefits from this
- [ ] Bundle dependencies (#67)
- [ ] Ensure that documentation / terminology is consistent across the project
	- Improve error messages - consistent format, make sure all cases are covered
- [ ] Look into lazy loading `cli.pkg` (#104)

LitoMore:
I don't really want that. You just reorganize the project to your liking.

sindresorhus:
The main reason I usually don't want TypeScript is that it introduces a build-step. If `- Bundle dependencies` ends up requiring a build-step, it may make sense to move to TypeScript.

sindresorhus:
@tommy-mitchell Should we release a new major version or do you intend to work on more items here?

tommy-mitchell:
@sindresorhus sorry for the late response, I’ve been busy with work this past week. Probably not going to have much more time for another week or two, so I think we could release now and add e.g. subcommands in a minor version

tommy-mitchell:
@sindresorhus thoughts on moving to TypeScript, now that we have a build step? We can use Rollup to emit the built JavaScript, and since a build is generated before testing, we can use `tsd` on the built `index.d.ts`.

sindresorhus:
Closing as everything on this list is fixed, except TS. And I'm not sure I will do it now.

PR #252, opened 2024-01-28
Rewrite in TypeScript

tommy-mitchell:
Per #233.

- Augmented the types from `minimist-options` to support the `arguments` option (see [`minimist-options.d.ts`](https://github.com/tommy-mitchell/meow/blob/940938be4d7b2b38c8b810408bd3e16d8ace8d47/source/minimist-options.d.ts#L47-L59))
	- I wanted to upstream these changes, but I ran into microsoft/TypeScript#17867 ([SO thread](https://stackoverflow.com/questions/61431397/how-to-define-typescript-type-as-a-dictionary-of-strings-but-with-one-numeric-i))
	- Not sure why it's working here but not when I make the changes to `minimist-options` itself
- Related, I kept running into that issue with other types as well, such as the `--` option in `flags`


readme.md
```diff
@@ -292,6 +292,7 @@ Default: `true`
 
 Whether to allow unknown flags or not.
 
+<!-- TODO: can this be removed now that hardRejection is deprecated? -->
 ## Promises
 
 Meow will make unhandled rejected promises [fail hard](https://github.com/sindresorhus/hard-rejection) instead of the default silent fail. Meaning you don't have to manually `.catch()` promises used in your CLI.
```
tommy-mitchell, line 298: See TODO comment

source/index.ts
```diff
@@ -0,0 +1,152 @@
...
+		process.exit(0);
+	};
+
+	if (argv._.length === 0 && options.argv.length === 1) {
+		if (argv['version'] === true && options.autoVersion) {
+			showVersion();
+		} else if (argv['help'] === true && options.autoHelp) {
+			showHelp(0); // TODO: why is this 0 and showHelp is 2?
+		}
+	}
```
tommy-mitchell, line 52: Why does `cli.showHelp()` exit with code 2, but the auto help exits with code 0?

source/index.ts
```diff
@@ -0,0 +1,152 @@
...
+//	flags: {rainbow: true},
+//	...
+//}
+
+foo(cli.input.at(0), cli.flags);
+```
+*/
+export default function meow<Flags extends AnyFlags>(helpMessage: string, options?: Options<Flags>): Result<Flags>;
+export default function meow<Flags extends AnyFlags>(options?: Options<Flags>): Result<Flags>;
+// TODO: should these be optional? importMeta is always needed
```
tommy-mitchell, line 133: Does it make sense for the `options` parameter to be optional? `meow` will just throw anyway without it, since `options.importMeta` is undefined in that case

source/options.ts
```diff
@@ -68,16 +75,22 @@ export const buildOptions = (helpText, options) => {
...
+	const parsedOptions: ParsedOptions = {
+		pkg,
 		argv: process.argv.slice(2),
 		flags: {},
 		inferType: false,
 		input: 'string',
-		help: helpText,
+		description: pkg.description ?? false,
+		help: helpMessage,
+		version: pkg.version ?? '', // TODO: maybe say "No version found"?
```
tommy-mitchell, line 89: Should we add a fallback for the version message? Currently, without `options.version` and `package.json.version`, the result is empty (tested by removing `?? ''`):

```sh
foo-cli/
$ foo --version

foo-cli/
$
```

test/test.ts
```diff
@@ -21,40 +20,32 @@ test('return object', t => {
 		flags: {
 			unicorn: {shortFlag: 'u'},
 			meow: {default: 'dog'},
-			'--': true,
+			'--': true as unknown as {type: 'string'; isMultiple: true},
+			// TODO: can not model due to microsoft/TypeScript#17867
+			// A potential workaround is to use a symbol for '--' -> {flags: [argvDashDash]: true}
```
tommy-mitchell, line 25: One of the "rest index type" issues

test/test.ts
```diff
@@ -299,20 +333,17 @@ test('supports `number` flag type - no flag but default', t => {
...
-			const {stdout} = await execa('node', ['estest/index.js', '--version'], {
-				importMeta: path.join(__dirname, '..'),
-			});
-			t.regex(stdout, /1.2.3/);
-		} catch (error) {
-			t.is(error, undefined);
-		}
+		const {stdout} = await spawnFixture('estest/index.ts', ['--version']);
+		t.regex(stdout, /1.2.3/);
 	});
```
tommy-mitchell, line 348: Is this test needed anymore?

tommy-mitchell:
May have missed some comments, but that's the bulk of where I'm needing clarification. The other TODOs I have are reminders to fix type issues.

source/index.ts
sindresorhus: Because `cli.showHelp()` is generally used to show the help on command misuse / incorrect command. While normal `cli --help` is not incorrect.

https://github.com/sindresorhus/meow/pull/47#issuecomment-244699098

This could use a code comment though, indeed.

source/index.ts
sindresorhus: It should not be optional.

test/test.ts
sindresorhus: Can be removed

test/test.ts
sindresorhus: Use a proper URL over `microsoft/TypeScript#17867`

source/options.ts
sindresorhus: Yeah, it can be `No version found`.

readme.md
tommy-mitchell: https://github.com/sindresorhus/meow/blob/fd0bc62ce47781e11da506b8e38e8668eb78a584/readme.md?plain=1#L16

here too

sindresorhus:
FYI: https://github.com/sindresorhus/meow/commit/e9a55cdd748d3f451c761c46eba1cfb7ca862dbe

tommy-mitchell:
Do we support a shorthand for setting flags?

```js
const cli = meow({
	importMeta: import.meta
	flags: {
		rainbow: 'boolean',
		unicorn: 'string',
	},
});
```

tommy-mitchell:
Got it. I think this test case (and maybe another?) had confused me before and I was remembering something about it:

https://github.com/sindresorhus/meow/blob/c97c03161753f611798996978e8274323c004382/test/errors.js#L38-L40

sindresorhus:
That test case is incorrect.

tommy-mitchell:
Updated to reflect changes from other PRs. Type tests are failing due to `tsd` not seeing `minimist-options.d.ts`.

source/options.ts
```diff
@@ -73,14 +84,14 @@ export const buildOptions = (helpText, options) => {
 	const pkg = foundPackage ?? {};
 	normalizePackageData(pkg);
 
-	const parsedOptions = {
+	const parsedOptions: ParsedOptions = {
 		argv: process.argv.slice(2),
 		flags: {},
 		inferType: false,
-		input: 'string',
+		input: 'string', // TODO: undocumented option?
```
tommy-mitchell: Should this option be documented?

sindresorhus: No, it doesn't really make sense to define a different type for all positional arguments.

tommy-mitchell:
CI is failing due to https://github.com/sindresorhus/type-fest/pull/824

tommy-mitchell:
I'm not sure how to fix the timeout issue, I get it sometimes locally when running `npm test` but never with `npx ava`.

sindresorhus:
Maybe try downgrading to `ava@5` for now.

tommy-mitchell:
Didn't work, guess it's something on our end. Now it doesn't even say which test timed out.

sindresorhus:
Works for me now: https://github.com/sindresorhus/meow/pull/252/commits/64f0915b8aa0ed42b9a04fc4bc7b814e125b57d3

sindresorhus:
Hmm, or maybe not. Maybe the problem here is `tsimp`. I never had any problems with `tsx`. Maybe worth trying that.

sindresorhus:
@tommy-mitchell Is this something you still intend to finish or should I close?

PR #253, opened 2024-01-31
add `types` to package.json

mmkal:
I'm surprised this is needed, but it seems to be. Types are not being inferred with just the `exports.types` field. This seems to be the case on npmjs.com, in VSCode, and via the typescript compiler.

Adding `types` seems to fix.

A few ways to see it's not right at the moment:

[Demo in typescript playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBCAphA7nAZlCI4HImq4BQRoksCyKAkgHYyJS0CGANhljvlQPQBGAV2CsAJj2C0RiAB4A6AFYBnYkA) (note that if you reach into the package via `import meow from 'meow/build/index.js'`, you do get the correct types 🤷.

[Are The Types Wrong result](https://arethetypeswrong.github.io/?p=meow%4013.1.0)

[`DT` icon on npmjs.com](https://www.npmjs.com/package/meow/v/13.1.0):

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
 		"types": "./build/index.d.ts",
 		"default": "./build/index.js"
 	},
+	"types": "./build/index.d.ts",
 	"sideEffects": false,
 	"engines": {
 		"node": ">=18"


sindresorhus:
The current config is correct.

---

> This seems to be the case on npmjs.com, in VSCode, and via the typescript compiler.

Then you should open an issue in those places. Usually, this issue is caused by incorrect tsconfig (not having `module` and `moduleResolution` set to `node16`.

mmkal:
I'll see if there's an existing issue in typescript. In the meantime, I came across this on `np`, working on https://github.com/sindresorhus/np/pull/730, which is written in plain js. So no tsconfig at all. The inferred type for meow in vscode is `any`. Would you suggest jsconfig.json there?

sindresorhus:
> So no tsconfig at all. The inferred type for meow in vscode is any. Would you suggest jsconfig.json there?

No, I would argue this is a VSCode/TS issue for when using comment annotations. It should respect that `meow` is a ESM package with an `exports` field and use `node16` resolution.

mmkal:
You're right. Even so, adding a jsconfig there/changing package.json here are much quicker quality-of-life improvements. They can be done today.

sindresorhus:
> Even so, adding a jsconfig there/changing package.json here are much quicker quality-of-life improvements. They can be done today.

I'm willing to add a jsconfig to `np` if you open the mentioned issues. I want to make sure the jsconfig can be removed eventually.

mmkal:
Will do. Combing through existing issues first - "exports" and "types" don't narrow down typescript's 5k open issues much!

Worth noting, I'm fine to do it but I still think this temporary solution is better than jsconfig.json. It's a one line fix that solves for all js projects that use meow. A jsconfig there is a new file that only solves for one repo.

sindresorhus:
> Worth noting, I'm fine to do it but I still think this temporary solution is better than jsconfig.json. It's a one line fix that solves for all js projects that use meow. A jsconfig there is a new file that only solves for one repo.

Yes, but it's unknown whether that could potentially cause problems with other tools. And if everyone simply works around TS bugs, they have no incentive to fix things.

PR #255, opened 2024-02-28
Improve tests

tommy-mitchell:
Separating some test modernization out from #252. Need to do a once over to check everything but I'm pretty confident with this.

Tried renaming files to keep blame history, but not all of them made it through. Oh well.


test/options/pkg.js
```diff
@@ -0,0 +1,57 @@
...
+
+	t.like(cli, {
+		pkg: {
+			name: 'browser-sync',
+			version: '',
+		},
+	});
+
+	// TODO: test that showVersion logs undefined
+});
```
tommy-mitchell, line 42: I think these test cases can be added later.

PR #256, opened 2024-02-28
Clean up `description` parsing

tommy-mitchell:
Separating out from #252.

Probably shouldn't have written this to begin with:

```js
description &&= help ? redent(`\n${description}\n`, options.helpIndent) : `\n${description}`;
help = `${description || ''}${help}\n`;
```

### DIFF
--- a/source/index.js
+++ b/source/index.js
@@ -3,13 +3,11 @@ import parseArguments from 'yargs-parser';
 import camelCaseKeys from 'camelcase-keys';
 import {trimNewlines} from 'trim-newlines';
 import redent from 'redent';
-import normalizePackageData from 'normalize-package-data';
 import {buildOptions} from './options.js';
 import {buildParserOptions} from './parser.js';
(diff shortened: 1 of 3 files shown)


source/index.js
```diff
@@ -76,7 +78,7 @@ const buildResult = (options, parserOptions) => {
 		input,
 		flags,
 		unnormalizedFlags,
-		pkg: package_,
+		pkg: packageJson,
```
tommy-mitchell, line 81: Thoughts on renaming `cli.pkg` to `package` or `packageJson`?

sindresorhus: It cannot be `package` as it's a reserved keyword, so it cannot be destructured.

tommy-mitchell: Hadn't considered the destructuring part, fair point. I'm ambivalent on the change, but feel free to add it to #233 if you'd like to change it (or just go for it). The consistency (and not having to rename variables) would be nice though.

PR #258, opened 2024-02-29
Remove `false` from `version` type, add fallback message

tommy-mitchell:
`false` has had no effect since #68.

---

Splitting out from #252 (ref: https://github.com/sindresorhus/meow/pull/252#discussion_r1468905743).

### DIFF
--- a/source/options.js
+++ b/source/options.js
@@ -75,6 +75,7 @@ export const buildOptions = (helpText, options) => {
 		inferType: false,
 		input: 'string',
 		help: helpText,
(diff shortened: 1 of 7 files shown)


source/options.js
```diff
@@ -75,6 +75,7 @@ export const buildOptions = (helpText, options) => {
 		inferType: false,
 		input: 'string',
 		help: helpText,
+		version: foundPackage?.packageJson.version || 'No version found',
```
tommy-mitchell, line 78: Using `||` over `??` because in #256, the normalization turns `pkg.version` into an empty string.

PR #266, opened 2026-06-13
Fix TypeScript runtime contract mismatches

dbrowning2:
## Summary
- Type positional input according to explicit input options and inferType.
- Type isRequired callbacks as receiving parsed runtime values instead of flag definitions.
- Reject dashed flag keys in the declaration tests to match runtime validation.

## Validation
- npm test


sindresorhus:
This makes the positional boolean input type stricter than what the runtime actually returns.

I manually checked `input: 'boolean'` and `input: 'boolean-array'` with `argv: ['true', 'false']`, and both return `['true', 'false']` where each value is still a string. With this change, TypeScript says `cli.input` is `boolean[]`, so user code can compile while being wrong at runtime.

I think this should either keep boolean positional input typed as `string[]`, or the runtime needs to actually parse these positional values as booleans. The first option seems simpler unless boolean positional parsing is inteded to be supported.

dbrowning2:
My apologies, I had AI working on a private project that has the same name (completely unrelated) to your project and it went off the rails, Claude Opus 4.8 if you're curious.  Closing this and adding some gates.  Sorry about that.

